Issue #1, opened 2022-10-14
audit-scanner - MVP Stage 1 - find relevant Policies for a given Namespace


flavio:
Implement a cli program, written in Go for its kube API filtering features.
Acceptance criteria:
  - The program takes as input the name of the Namespace to be inspected
  - Given the namespace, it will return a list of policies (both `ClusterAdmissionPolicy` and `AdmissionPolicy`) that are interested about this namespace
  - Only active policies are considered
  - Policies targeting `*` are not considered
  - Only policies interested about `CREATE` events are considered
  - The list is printed to stdout
  - Debug/error messages should be handled with a Go logging framework to be decided. Structured logging should be used
  - We have unit tests
  - Have integration tests **if really needed**

PR #2, opened 2022-10-25
Audit scanner MVP1

raulcabello:
## Description

The program takes as input the name of the `Namespace` to be inspected, then print the policies that will be audited.

```
Usage:
  audit-scanner [flags]

Flags:
  -h, --help               help for audit-scanner
  -l, --loglevel string    level of the logs. Supported values are: [trace debug info warn error fatal] (default "info")
  -n, --namespace string   namespace to be evaluated
```

It is using [zerolog](https://github.com/rs/zerolog) for the structured logging. We can easily change it if you prefer to use another library. I also considered using [zap](https://pkg.go.dev/go.uber.org/zap), but I find zerolog easier to work with. Let me know what you think

## Test
- run `make` to create the binary. It will use the `kubeconfig` in your home dir if it is not deployed in a Kubernetes cluster.
Make sure you have the latest CRD from the [feat-audit](https://github.com/kubewarden/kubewarden-controller/tree/feat-audit) branch deployed in your cluster, otherwise `backgroundAudit` will always be **false**.
- deploy `ClusterAdmissionPolicies` and `AdmissionPolicies` in your cluster
- run `./audit-scanner -n namespace -l trace ` loglevel (-l) is optional, but it is useful to set it to trace when testing. It will show which policies were skipped from being audited, and the reason.
- unit-tests can be run with `make test`

Fix https://github.com/kubewarden/audit-scanner/issues/1


internal/client/client.go
```diff
@@ -0,0 +1,161 @@
...
+		return nil, err
+	}
+
+	return newClient(config)
+}
+
+func newClient(config *rest.Config) (Client, error) {
+	// register Kubewarden policies in scheme
+	customScheme := scheme.Scheme
+	customScheme.AddKnownTypes(schema.GroupVersion{Group: kubewardenPoliciesGroup, Version: kubewardenPoliciesVersion}, &policiesv1.ClusterAdmissionPolicy{}, &policiesv1.AdmissionPolicy{}, &policiesv1.ClusterAdmissionPolicyList{}, &policiesv1.AdmissionPolicyList{})
```
viccuad: we don't need to register v1alpha2 policies, correct? Because they have been migrated?

internal/policies/fetcher.go
```diff
@@ -0,0 +1,127 @@
...
+		return nil, err
+	}
+	policies := &policiesv1.ClusterAdmissionPolicyList{}
+	err = f.client.List(context.Background(), policies, &client.ListOptions{})
+	if err != nil {
+		return nil, fmt.Errorf("can't list AdmissionPolicies: %v", err)
+	}
+
+	for _, policy := range policies.Items {
+		policy := policy
```
viccuad, line 88:
```suggestion
```

viccuad: I think this isn't needed. We don't intend to update `policy` meanwhile we are going through the range.

viccuad (approved):
LGTM! just some silly questions.

internal/client/client.go
raulcabello: yes

internal/policies/fetcher.go
raulcabello: yes, we need this. We [append](https://github.com/kubewarden/audit-scanner/blob/a7a141fcfd8d91b287ec4d849f4586e241030e81/internal/policies/fetcher.go#L94) `&policy` which will be updated in each loop iteration

jvanz (approved):

internal/policies/fetcher.go
viccuad: doh! :sweat_smile: . Resolving!

Issue #36, opened 2023-04-26
audit-scanner - MVP Stage 4 - create PolicyReport


jvanz:
Implement creation of `PolicyReport` inside of the namespace being analyzed, by using the kube API client and [https://github.com/kubernetes-sigs/wg-policy-prototypes](https://github.com/kubernetes-sigs/wg-policy-prototypes).

Acceptance criteria:

-   See [PolicyReport RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0012-policy-report.md) and [audit RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0011-audit-checks.md).
-   We have unit tests
-   We have integration test **if needed**
-   Take care of errors if PolicyReports CRDs are not present in cluster.

jvanz:
@kubewarden/kubewarden-developers , I'm thinking about a conversation that I had with Flavio about storing the results on etcd (Policy Report). As it is planned now. During this conversation Flavio mentioned he has talked with a SUSE colleague and discuss the scalability issues that can impose on us 

The first idea would be move away from etcd and use another database, like sqlite. I think this is fine. But we should abstract the backend database from the scanner. Therefore, we can easily extend to other types of databases on user request. But there is a thing that we need to reformulate. How users will access this information?

As we move away from the policy report format, we need to allow users to easily access this information. One option is using OTEL, Prometheus and Grafana or something similar. The good point on that is that we get a nice UI out of the box. I still need to refresh my memory to see what's the available metrics types and if they will be enough to cover the functionality that we planned to have using Policy Report. The bad side is that user must have this stack installed to visualize the information (which I think is pretty bad)

Another option is create an API to expose this to the outer world. If we do that now, we need to think of scalability again. Because the API must be up and running mos of the time. Furthermore, we need to think of a format to expose this data as well. 

If we decided to go with the OTEL path, we can use the sqlite database as [the cache for future evaluations](https://github.com/kubewarden/rfc/blob/main/rfc/0011-audit-checks.md#step-5-attempt-to-reuse-previous-evaluation-results) and export the final result in the metrics. Another thing to remember is that with a sqlite database in the picture we start to worry about database migrations (e.g. how to add/remove a new field?, etc)

tldr: what do you think of using OTEL, Prometheus and Grafana stack to expose audit scanner reports?

Sorry for this random ideas. I would like to start the discussion with the team while I was thinking this as well.

viccuad:
[...]

PR #40, opened 2023-05-26
feat: Add report package, print reports to stdout

viccuad:
## Description

Partial implementation of https://github.com/kubewarden/audit-scanner/issues/36.

feat: Add report package, print reports to stdout

## Test

Manual for now.

Example:

<details>

<summary>output</summary>

```console
$ ./bin/audit-scanner -k kubewarden --namespace default -l error | jq .
{
  "metadata": {
    "name": "polr-clusterwide",
    "creationTimestamp": null
  },
  "summary": {
    "pass": 0,
    "fail": 0,
    "warn": 0,
    "error": 0,
    "skip": 0
  }
}
{
  "metadata": {
    "name": "polr-ns-default",
    "namespace": "default",
    "creationTimestamp": "2023-05-25T14:27:40Z"
  },
  "scope": {},
  "summary": {
    "pass": 5,
    "fail": 1,
    "warn": 0,
    "error": 0,
    "skip": 0
  },
  "results": [
    {
      "source": "kubewarden",
      "policy": "cap-drop-capabilities",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "pass"
    },
    {
      "source": "kubewarden",
      "policy": "cap-no-privilege-escalation",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "pass"
    },
    {
      "source": "kubewarden",
      "policy": "cap-no-privileged-pod",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "fail",
      "message": "Privileged container is not allowed"
    },
    {
      "source": "kubewarden",
      "policy": "cap-do-not-run-as-root",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "pass"
    },
    {
      "source": "kubewarden",
      "policy": "cap-no-host-namespace-sharing",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "pass"
    },
    {
      "source": "kubewarden",
      "policy": "cap-do-not-share-host-paths",
      "timestamp": {
        "seconds": 1685024860,
        "nanos": 0
      },
      "result": "pass"
    }
  ]
}
```

</details>

### Potential improvement

Missing:
- Add tests for report.Save().


internal/report/create.go
```diff
@@ -0,0 +1,140 @@
...
+	log.Debug().Any("resource", resource)
+
+	return &v1beta1.PolicyReportResult{
+		Source: PolicyReportSource,
+		Policy: name, // either cap-policy_name or ap-policy_name
+		// Timestamp shouldn't be used in go structs, and only gives seconds
+		// https://github.com/kubernetes/apimachinery/blob/v0.27.2/pkg/apis/meta/v1/time_proto.go#LL48C9-L48C9
+		Timestamp:   timestamp,               // time the result was found
+		Result:      result,                  // pass, fail, error
+		Subjects:    []*v1.ObjectReference{}, // TODO reference to object evaluated
```
viccuad, line 137: This is missing.

internal/report/create.go
```diff
@@ -0,0 +1,140 @@
...
+func New(namespace string, skippedNum int) error {
+	// policy report doesn't exist and needs to be created
+	if policyReports[namespace] == nil {
+		policyReports[namespace] = &PolicyReport{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:              "polr-ns-" + namespace,
+				Namespace:         namespace,
+				CreationTimestamp: metav1.Now(),
+			},
+			Scope: &v1.ObjectReference{}, // TODO to namespace
```
viccuad, line 52: this is missing; we have the `namespace` string. Ideally, report.New() should receive the Object already. Other backends (SQL) would be ok with a string I suppose.

internal/report/create.go
```diff
@@ -0,0 +1,140 @@
...
+		Fail:  0, // count of policies with requirements not met
+		Warn:  0, // not used for now
+		Error: 0, // count of policies that couldn't be evaluated
+		Skip:  0, // count of policies that were not selected for evaluation
+	},
+	Results: []*v1beta1.PolicyReportResult{},
+}
+
+// slice of namespaced PolicyReports
+var policyReports = make(map[string]*PolicyReport)
```
viccuad, line 38: This could be a store struct with a mutex, accompanied with a policyReportStore interface with Add(), Update(), Get(), etc.

internal/report/create.go
jvanz: I've pushed a commit  proposing a fix for that. ;)

internal/report/create.go
viccuad: done in https://github.com/kubewarden/audit-scanner/pull/40/commits/bf47e0dea867579da05812a3adf1215f555c225f

internal/report/create.go
jvanz: If we return a "report object"  to the scanner, we reduce the need to use mutexes as well. As the scanner run sequentially, there is no need to add a mutex there as well. Furthermore, different report objects can be used in different threads (gorotines). Where each one will scanning different namespaces with not concurrent access (depending on how we implement it, of cource). But I think  this is a something we can do later.

jvanz (approved):
Overall,  it looks good to me. I think we will need to change how we do some things. But I think we will notice what needs to be improved in future steps on the scanner. Therefore, I think we can merge it and iterate over it during the next steps.

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+	case schema.GroupVersionKind{
+		Group:   constants.KubewardenPoliciesGroup,
+		Version: constants.KubewardenPoliciesVersion,
+		Kind:    "ClusterAdmissionPolicy",
+	}:
+		name = "cap-" + policy.GetName()
+	case schema.GroupVersionKind{
+		Group:   constants.KubewardenPoliciesGroup,
+		Version: constants.KubewardenPoliciesVersion,
+		Kind:    "AdmissionPolicy",
```
flavio, line 123: Maybe this should become a constant too?

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+		policyReports[namespace].Summary.Pass++
+	}
+	policyReports[namespace].Results = append(policyReports[namespace].Results, result)
+}
+
+func newResult(policy policiesv1.Policy, resource unstructured.Unstructured, auditResponse *admv1.AdmissionReview, responseErr error) *v1beta1.PolicyReportResult {
+	var result v1beta1.PolicyResult
+	var description string
+	if responseErr != nil {
+		result = "error"
```
flavio, line 101: Don't they provide a constant for that?

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+		Version: constants.KubewardenPoliciesVersion,
+		Kind:    "AdmissionPolicy",
+	}:
+		name = "ap-" + policy.GetName()
+	}
+
+	time := metav1.Now()
+	timestamp := *time.ProtoTime()
+
+	log.Debug().Any("resource", resource)
```
flavio, line 131: I've not run the code, does this log message provide any clue about the context where it's being originated?

internal/report/write.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	polReport "sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1beta1"
+)
+
+// Write instantiates the passed namespaced PolicyReport if it doesn't exist, or
+// updated a new one if one is found
+func Write(report *polReport.PolicyReport) error {
```
flavio, line 20: nitpicking, maybe this should be named `Save` ?

internal/report/write.go
```diff
@@ -0,0 +1,81 @@
...
+	// Create new Policy Report if not found
+	if errors.IsNotFound(getErr) {
+		log.Info().Msg("creating policy report...")
+
+		err = client.Create(context.TODO(), report)
+		if err != nil {
+			return fmt.Errorf("failed when creating PolicyReport: %w", err)
+		}
+	} else {
+		// Update existing Policy Report
```
flavio, line 51: I wonder if we shouldn't just delete the old report and always create a new one :thinking:

internal/scanner/scanner.go
```diff
@@ -100,19 +109,29 @@ func (s *Scanner) ScanNamespace(namespace string) error {
...
+				Msg("resource under review")
 			admissionRequest := resources.GenerateAdmissionReview(resource)
-			auditResponse, err := sendAdmissionReviewToPolicyServer(url, admissionRequest, httpClient)
-			if err != nil {
-				// TODO what's the better thing to do here?
+			auditResponse, responseErr := sendAdmissionReviewToPolicyServer(url, admissionRequest, httpClient)
+			if responseErr != nil {
+				// log error, will end in PolicyReportResult too
 				log.Error().Err(err)
 				continue
```
flavio, line 136: The `continue` statement has to be removed, otherwise the `report.AddResult` function will not record the error.

Something like that:

```golang
if responseErr != nil {
  log.Error()
} else {
  // need to be inside of else, otherwise the code will panic when trying to access auditResponse.<something>
  log.Debug
}

report.AddResult
```

```

**Review by flavio — CHANGES_REQUESTED** [review]

**Comment by viccuad** [comment]
Force-pushed a new version (sorry), with partial history, since I refactored heavily everything.

File: internal/report/create.go (hunk 3 above)
Comment (viccuad) [reply internal/report/create.go]: resolving this, now it is a proper store

File: internal/constants/constants.go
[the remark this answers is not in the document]
Comment (viccuad) [reply internal/constants/constants.go:1]: sadly they aren't defined, so I added the missing ones.

File: internal/policies/fetcher.go [hunk 10]
```diff
@@ -39,26 +35,55 @@ func NewFetcher() (*Fetcher, error) {
 }
 
 // TODO implement this for all ns
-func (f *Fetcher) GetPoliciesForAllNamespaces() ([]policiesv1.Policy, error) {
-	return nil, errors.New("scanning all namespaces is not implemented yet. Please pass the --namespace flag to scan a namespace")
+func (f *Fetcher) GetPoliciesForAllNamespaces() ([]policiesv1.Policy, int, error) {
```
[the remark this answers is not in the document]
viccuad: Done

internal/policies/fetcher.go
```diff
@@ -39,26 +35,55 @@ func NewFetcher() (*Fetcher, error) {
...
 		namespacePolicies[namespace] = append(namespacePolicies[namespace], &policy)
 	}
 
-	return f.filter(namespacePolicies[namespace]), nil
+	filteredPolicies := f.filter(namespacePolicies[namespace])
+	skippedNum := len(namespacePolicies[namespace]) - len(filteredPolicies)
+	return filteredPolicies, nsReference, skippedNum, nil
+}
+
+func (f *Fetcher) getNamespaceReference(nsName string) (*v1.ObjectReference, error) {
```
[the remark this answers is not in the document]
viccuad: Done at https://github.com/kubewarden/audit-scanner/pull/40/files#diff-4064648b92b1e9eae66c02b7f4342c24955f50d5ad0aae3f0adf6e37881442ebR66-R79

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+
+// New() creates a new PolicyReport for the corresponding namespace if it
+// doesn't exist, setting the number of skipped policies in the
+// PolicyReport{}.Summary.
+func New(namespace *v1.ObjectReference, skippedNum int) error {
+	if policyReports[namespace.Name] == nil {
+		// policy report doesn't exist and needs to be created
+		policyReports[namespace.Name] = &PolicyReport{
+			ObjectMeta: metav1.ObjectMeta{
+				Name:              "polr-ns-" + namespace.Name,
```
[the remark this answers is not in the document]
viccuad: Here I just followed the RFC: https://github.com/kubewarden/rfc/blob/main/rfc/0012-policy-report.md#policyreport-detailed-overview
I think we went with that nomenclature as other projects have a similar or the same prefix.

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+}
+
+// slice of namespaced PolicyReports
+var policyReports = make(map[string]*PolicyReport)
+
+// New() creates a new PolicyReport for the corresponding namespace if it
+// doesn't exist, setting the number of skipped policies in the
+// PolicyReport{}.Summary.
+func New(namespace *v1.ObjectReference, skippedNum int) error {
+	if policyReports[namespace.Name] == nil {
```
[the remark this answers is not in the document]
viccuad: Done, moved into a store that [gets instantiated in Scanner{}](https://github.com/kubewarden/audit-scanner/pull/40/files#diff-95e0a286796962f784af743a32471dada7258c4789b441aaca0a77653e3c1155R45).

internal/report/create.go
viccuad: https://github.com/kubewarden/audit-scanner/pull/40/files#diff-02820ac94e512b02b1879746ff901479469e1dda3c21277d7f7a170cd1c25c71R6-R7

internal/report/create.go
viccuad: No, but added our own at https://github.com/kubewarden/audit-scanner/pull/40/files#diff-1bbce338ff2b1a40ca20a61d9962efd9cfbeb4a6c8ef9279f29c22733f3d5913R40-R47

internal/report/create.go
viccuad: This was a leftover debug message, removed now.

internal/scanner/scanner.go
viccuad: Woops, removed `continue` message.

internal/report/create.go
```diff
@@ -0,0 +1,152 @@
...
+				Error: 0,          // count of policies that couldn't be evaluated
+				Skip:  skippedNum, // count of policies that were not selected for evaluation
+			},
+			Results: []*v1beta1.PolicyReportResult{},
+		}
+	}
+	return nil
+}
+
+func Print() {
```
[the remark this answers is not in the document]
viccuad: Done. Created a muxed store at `report/store`, and provided `store.ToJSON()`

internal/report/write.go
viccuad: Yes, better. Renamed (now it is at internal/report/store.go).

internal/report/write.go
viccuad: I thought that naively deleting the old report without RetryOnConflict could be somewhat dangerous. And once I had the retry, why not just do an update for the full Report, it is going to overwrite it anyways. Plus this way, we bump the version for possible listeners.

CONTRIBUTING.md
```diff
@@ -0,0 +1,35 @@
...
+  --namespace kubewarden \
+  --create-namespace \
+  kubewarden-defaults kubewarden/kubewarden-defaults \
+  --set recommendedPolicies.enabled=True \
+  --set recommendedPolicies.defaultPolicyMode=monitor
+
+kubectl port-forward -n kubewarden service/policy-server-default 3000:8443
+```
+
+Then, patch  `GetPolicyServerURLRunningPolicy()` with:
```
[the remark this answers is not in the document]
viccuad: Agree, will add it in this or follow-up PR.

cmd/root.go
```diff
@@ -52,7 +52,11 @@ There will be a ClusterPolicyReport with results for cluster-wide resources.`,
 			}
 			scanner := scanner.NewScanner(policiesFetcher, resourcesFetcher)
 
-			return startScanner(namespace, scanner)
+			err = startScanner(namespace, scanner)
+			if err != nil {
+				return err
+			}
+			return nil
```
flavio, line 63: I think the original code was better (and idiomatic), this is just a more verbose way to do the very same thing.

internal/policies/fetcher.go
```diff
@@ -38,27 +35,48 @@ func NewFetcher() (*Fetcher, error) {
...
+	err := f.client.Get(context.Background(),
+		client.ObjectKey{
+			Name: nsName,
+		},
+		namespace)
+	if err != nil && errorsApi.IsNotFound(err) {
+		return nil, fmt.Errorf("namespace not found: %s", nsName)
+	}
+	if err != nil {
+		return nil, fmt.Errorf("can't get namespace: %s", nsName)
```
flavio, line 77: We should not hide the error, otherwise it will be hard to debug stuff

internal/report/client.go
flavio, line 1: this is a bit nitpicking, but I don't see the connection between the name of this file (`client.go`) and what is actually done inside of it. I think this should be moved inside of another file?

**TODO:** get back to that at the end of the review

internal/policies/fetcher.go
```diff
@@ -134,8 +152,8 @@ func (f *Fetcher) getAdmissionPolicies(namespace string) ([]policiesv1.Admission
 func newClient() (client.Client, error) {
 	config := ctrl.GetConfigOrDie()
 	customScheme := scheme.Scheme
-	customScheme.AddKnownTypes(schema.GroupVersion{Group: kubewardenPoliciesGroup, Version: kubewardenPoliciesVersion}, &policiesv1.ClusterAdmissionPolicy{}, &policiesv1.AdmissionPolicy{}, &policiesv1.ClusterAdmissionPolicyList{}, &policiesv1.AdmissionPolicyList{})
-	metav1.AddToGroupVersion(customScheme, schema.GroupVersion{Group: kubewardenPoliciesGroup, Version: kubewardenPoliciesVersion})
+	customScheme.AddKnownTypes(schema.GroupVersion{Group: constants.KubewardenPoliciesGroup, Version: constants.KubewardenPoliciesVersion}, &policiesv1.ClusterAdmissionPolicy{}, &policiesv1.AdmissionPolicy{}, &policiesv1.ClusterAdmissionPolicyList{}, &policiesv1.AdmissionPolicyList{})
```
flavio, line 155: nitpicking, can we indent a bit the code to make it more readable? like you have done inside of `report/client.go`

internal/report/client.go
```diff
@@ -0,0 +1,81 @@
...
+			report.SetResourceVersion(getObj.GetResourceVersion())
+
+			updateErr := client.Update(context.TODO(), report)
+			// return unwrapped error for RetryOnConflict()
+			return updateErr
+		})
+		if retryErr != nil {
+			log.Error().Err(retryErr).Msg("PolcyReport update failed")
+		}
+		log.Info().Msg("updated policy report")
```
flavio, line 78: Maybe add some details about the report, like the name and the namespace. Do that via structured logging please

internal/report/report.go
```diff
@@ -0,0 +1,183 @@
...
+// Status specifies state of a policy result
+const (
+	StatusPass  = "pass"
+	StatusFail  = "fail"
+	StatusWarn  = "warn"
+	StatusError = "error"
+	StatusSkip  = "skip"
+)
+
+func NewCPR(name string) *ClusterPolicyReport {
```
flavio, line 49: nitpicking: I would prefer a more explicit name: `NewClusterPolicyReport`

internal/report/report.go
```diff
@@ -0,0 +1,183 @@
...
+				Warn:  0, // not used for now
+				Error: 0, // count of policies that couldn't be evaluated
+				Skip:  0, // count of policies that were not selected for evaluation
+			},
+			Results: []*v1alpha2.PolicyReportResult{},
+		},
+	}
+}
+
+func NewPR(namespace *v1.Namespace) *PolicyReport {
```
flavio, line 68: nitpicking: I would prefer a more explicit name: `NewPolicyReport`

internal/report/report.go
```diff
@@ -0,0 +1,183 @@
...
+		r.Summary.Fail++
+	case StatusError:
+		r.Summary.Error++
+	case StatusPass:
+		r.Summary.Pass++
+	}
+	r.Results = append(r.Results, result)
+}
+
+func newResult(policy policiesv1.Policy, resource unstructured.Unstructured, auditResponse *admv1.AdmissionReview, responseErr error) *v1alpha2.PolicyReportResult {
```
flavio, line 130: nitpicking: `newPolicyReportResult`

internal/report/report.go
flavio, line 68: nitpicking: I would prefer a more explicit name: `NewPolicyReport`

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+type policyReportStore struct {
+	// store is a map of namespaces and namespaced PolicyReports
+	store map[string]RInterface
+	// clusterPR is the sole ClusterPolicyReport
+	clusterPR ClusterPolicyReport
+
+	rwm *sync.RWMutex
+}
+
+type PolicyReportStore interface {
```
flavio, line 19: Why do we need this interface?

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+)
+
+// PolicyReportStore caches the latest version of PolicyReports
+type policyReportStore struct {
+	// store is a map of namespaces and namespaced PolicyReports
+	store map[string]RInterface
+	// clusterPR is the sole ClusterPolicyReport
+	clusterPR ClusterPolicyReport
+
+	rwm *sync.RWMutex
```
flavio, line 16: I would go with something like that:

```golang
namespacePolicyReports: make(map[string]PolicyReport)
clusterPolicyReport ClusterPolicyRepor

namespacePolicyReportsMutex *sync.RWMutex
clusterPolicyReportsMutex *sync.RWMutex
```

Also, I would remove the `RInterface`. I don't see (yet) why we should benefit from that

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+// clusterwide ClusterPolicyReport
+func NewPolicyReportStore() PolicyReportStore {
+	return &policyReportStore{
+		store:     make(map[string]RInterface),
+		clusterPR: *NewCPR("clusterwide"),
+		rwm:       new(sync.RWMutex),
+	}
+}
+
+func (s *policyReportStore) Add(report RInterface) error {
```
flavio, line 47: Make that a `AddPolicyReport` and then introduce a new `AddClusterPolicyReport`

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	if report.GetType() == ClusterPolicyReportType {
+		cpr, _ := report.(*ClusterPolicyReport)
+		s.clusterPR = *cpr
+	} else {
+		s.store[report.GetNamespace()] = report
+	}
+	return nil
+}
+
+func (s *policyReportStore) Get(namespace string) (RInterface, error) {
```
flavio, line 59: Rename that to `GetPolicyReport`, return a `PolicyReport`

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	s.rwm.RLock()
+	report, present := s.store[namespace]
+	s.rwm.RUnlock()
```
flavio, line 62: you can also use a `defer`, you can take as many read locks as you want, hence we don't have to give it back immediately

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	if present {
+		return report, nil
+	}
+	return nil, errors.New("report not found")
+}
+
+func (s *policyReportStore) GetClusterWide() (RInterface, error) {
```
flavio, line 69: rename to `GetClusterPolicyReport` and drop the `RInterface`

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	s.rwm.RLock()
+	report := s.clusterPR
+	s.rwm.RUnlock()
```
flavio, line 72: same here, we can use `defer`

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	return RInterface(&report), nil
+}
+
+func (s *policyReportStore) Update(report RInterface) error {
```
flavio, line 76: same change here, differentiate between namespace and cluster wide

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+	if report.GetType() == ClusterPolicyReportType {
+		cpr, _ := report.(*ClusterPolicyReport)
+		s.clusterPR = *cpr
+	} else {
+		s.store[report.GetNamespace()] = report
+	}
+	return nil
+}
+
+func (s *policyReportStore) Remove(namespace string) error {
```
flavio, line 88: same here

internal/report/store.go
```diff
@@ -0,0 +1,121 @@
...
+		defer s.rwm.Unlock()
+		delete(s.store, namespace)
+	}
+	return nil
+}
+
+func (s *policyReportStore) RemoveAllNamespaced() error {
+	s.rwm.Lock()
+	defer s.rwm.Unlock()
+	s.store = make(map[string]RInterface)
```
flavio, line 100: Add a TODO, once 1.21 is out we should instead use the new `clear` builtin:
* PR: https://github.com/golang/go/issues/56351
* Docs: https://go-review.googlesource.com/c/go/+/467755/8/doc/go_spec.html#7186

internal/report/client.go
```diff
@@ -0,0 +1,81 @@
...
+	"k8s.io/client-go/kubernetes/scheme"
+	"k8s.io/client-go/util/retry"
+	ctrl "sigs.k8s.io/controller-runtime"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	polReport "sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1beta1"
+)
+
+// Save instantiates the passed namespaced PolicyReport if it doesn't exist, or
+// updated a new one if one is found
+func Save(report *polReport.PolicyReport) error {
```
flavio, line 20: I think this should be moved into the store object. It would be more clear and also better, here we allocate a client and configure that for each policy report that we need to write; doing that inside of the Store would require just one single instance

internal/resources/fetcher.go
```diff
@@ -25,6 +27,9 @@ type Fetcher struct {
 	// Namespace where the Kubewarden components (e.g. policy server) are installed
 	// This is the namespace used to fetch the policy server resources
 	kubewardenNamespace string
+	// FQDN of the policy server to query. If not empty, Fetcher will query on
+	// port 3000. Useful for out-of-cluster debugging
```
flavio, line 31:
```suggestion
	// `localhost:3000`. Useful for out-of-cluster debugging
```

internal/resources/fetcher.go
```diff
@@ -38,11 +43,13 @@ type AuditableResources struct {
 }
 
 // NewFetcher returns a new fetcher with a dynamic client
-func NewFetcher(kubewardenNamespace string) (*Fetcher, error) {
+func NewFetcher(kubewardenNamespace string, policyServerFQDN string) (*Fetcher, error) {
 	config := ctrl.GetConfigOrDie()
 	dynamicClient := dynamic.NewForConfigOrDie(config)
-
-	return &Fetcher{dynamicClient, kubewardenNamespace}, nil
+	if policyServerFQDN != "" {
```
flavio, line 49: I think we should assume `policyServerFQDN` is not empty. We should have cobra set a default value when the user does not provide it

internal/scanner/scanner.go
```diff
@@ -110,17 +109,19 @@ func auditResource(resource *resources.AuditableResources, resourcesFetcher *Res
...
+			if responseErr != nil {
+				// log error, will end in PolicyReportResult too
+				log.Error().Err(responseErr)
 				continue
 			}
 
 			log.Debug().Dict("response", zerolog.Dict().
 				Str("uid", string(auditResponse.Response.UID)).
 				Bool("allowed", auditResponse.Response.Allowed)).
 				Msg("audit review response")
```
flavio, line 122: wrap that into an `else { }` block, otherwise things will go south when `responseErr != nil`

flavio (commented):
nice refactoring, I left some comments :clap: :+1:

cmd/root.go
viccuad: was this about `Execute()`? I can't see any change on `startScanner()`. Maybe I'm missing something. Just in case, left `Execute()` back how it was.

internal/policies/fetcher.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/2db512b545e1262c1288a96bc55252ee2ac676ca.

internal/report/client.go
viccuad: Move to store.go in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee

internal/policies/fetcher.go
viccuad: done in https://github.com/kubewarden/audit-scanner/pull/40/commits/2db512b545e1262c1288a96bc55252ee2ac676ca

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/2db512b545e1262c1288a96bc55252ee2ac676ca

internal/report/client.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee

internal/report/client.go
viccuad: done in https://github.com/kubewarden/audit-scanner/pull/40/commits/2db512b545e1262c1288a96bc55252ee2ac676ca

internal/report/report.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/898d19140dc59093af5165a216f4595643d2e004

internal/report/report.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/898d19140dc59093af5165a216f4595643d2e004

internal/report/report.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/898d19140dc59093af5165a216f4595643d2e004

internal/report/report.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/898d19140dc59093af5165a216f4595643d2e004

internal/report/store.go
viccuad: That's better indeed. Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/report/store.go
viccuad: Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/0329296a7ee8c77e295d16a18d64d32492974eee.

internal/scanner/scanner.go
viccuad: Indeed. Done in https://github.com/kubewarden/audit-scanner/pull/40/commits/2db512b545e1262c1288a96bc55252ee2ac676ca.

internal/resources/fetcher.go
viccuad: Thing is, for knowing the FQDN we need to know the service of the PolicyServer pod, hence why it happens later.

We want the in-cluster behaviour as default, so that's why I left it empty, If we want our own (`localhost`, `cluster.local`, whatever the developer uses), we give it value with the flag.

internal/resources/fetcher.go
viccuad: Thing is, for knowing the FQDN we need to know the service of the PolicyServer pod, hence why it happens later.

We want the in-cluster behaviour as default, so that's why I left it empty, If we want our own (`localhost`, `cluster.local`, whatever the developer uses), we give it value with the flag.

viccuad:
I don't get why the unit tests are failing on CI, they pass locally. Will have a look at that next day.

cmd/root.go
flavio: I mean that `startScanner` is already returning an `error`, hence it's feels more natural to do:

```golang
return startStanner(namespace, scanner)
```

rather than:

```golang
err = startScanner(namespace, scanner)
if err != nil {
  return err
}
return nil
```

internal/scanner/scanner.go
```diff
@@ -36,71 +42,69 @@ type ResourcesFetcher interface {
...
 	}
-	log.Debug().Str("namespace", namespace).Int("count", len(policies)).Msg("number of policies to evaluate")
-
-	// TODO continue with the scanning and remove this code
-	log.Debug().Str("namespace", namespace).Msg("The following policies were found for the namespace " + namespace)
-	for _, policy := range policies {
-		log.Debug().Str("policy name", policy.GetName()).Msg("Policy retrieved")
+	namespace, err := s.policiesFetcher.GetNamespace(nsName)
+	if err != nil {
+		return err
```
flavio, line 69: nitpicking, maybe we should look for the `Namespace` first, then look for the other resources defined inside of it

flavio (commented):
I left some minor comments

internal/resources/fetcher.go
flavio: I see, thanks

internal/scanner/scanner.go
viccuad: True, done in https://github.com/kubewarden/audit-scanner/pull/40/commits/5fc30be0cad27129eda6b3c4a38e10a336d5cc22.

internal/policies/fetcher.go
```diff
@@ -38,27 +35,48 @@ func NewFetcher() (*Fetcher, error) {
...
+
+func (f *Fetcher) GetNamespace(nsName string) (*v1.Namespace, error) {
+	namespace := &v1.Namespace{}
+	err := f.client.Get(context.Background(),
+		client.ObjectKey{
+			Name: nsName,
+		},
+		namespace)
+	if err != nil && errorsApi.IsNotFound(err) {
+		return nil, fmt.Errorf("namespace %s not found: %w", nsName, err)
```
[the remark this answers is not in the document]
viccuad: True, done in https://github.com/kubewarden/audit-scanner/pull/40/commits/5fc30be0cad27129eda6b3c4a38e10a336d5cc22.

cmd/root.go
```diff
@@ -83,5 +94,6 @@ func startScanner(namespace string, scanner Scanner) error {
 func init() {
 	rootCmd.Flags().StringP("namespace", "n", "", "namespace to be evaluated")
 	rootCmd.Flags().StringP("kubewarden-namespace", "k", defaultKubewardenNamespace, "namespace where the Kubewarden components (e.g. Policy Server) are installed (required)")
+	rootCmd.Flags().StringP("policy-server-fqdn", "p", "", "FQDN of the PolicyServers. If non-empty, they will be queried on <FQDN>:3000. Useful for out-of-cluster debugging")
```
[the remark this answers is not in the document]
viccuad: That is better indeed, done in https://github.com/kubewarden/audit-scanner/pull/40/commits/9b9835c9287a320a3cd911ea9556b041df65305c.

cmd/root.go
```diff
@@ -42,30 +43,40 @@ There will be a ClusterPolicyReport with results for cluster-wide resources.`,
...
 )
 
 // Execute adds all child commands to the root command and sets flags appropriately.
 // This is called by main.main(). It only needs to happen once to the rootCmd.
 func Execute() {
-	err := rootCmd.Execute()
-	if err != nil {
+	if err := rootCmd.Execute(); err != nil {
+		log.Fatal().Err(err).Msg("Error on cmd.Execute()")
 		os.Exit(1)
```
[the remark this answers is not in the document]
viccuad: True, done in https://github.com/kubewarden/audit-scanner/commit/5fc30be0cad27129eda6b3c4a38e10a336d5cc22.

viccuad:
I'm still puzzled why the unit-tests fail without any error for the report package. They pass locally, and I'm running the same go version. The go env vars also look fine.

flavio (approved):
thanks for all the changes, LGTM

cmd/root.go
viccuad: oh, indeed, done in https://github.com/kubewarden/audit-scanner/commit/5fc30be0cad27129eda6b3c4a38e10a336d5cc22.

CONTRIBUTING.md
jvanz, line 1: Thanks for this docs! Pretty useful!

jvanz (approved):
LGTM! Thanks for the patience! :)

PR #47, opened 2023-06-07
feat: Set PolicyReportResult fields

viccuad:
## Description

Relates to https://github.com/kubewarden/helm-charts/issues/237

- chore: Provide more info logs
- ci: Test container image on build
- feat: Add ManagedBy label
- Set PolicyReportResult fields:
  - severity: `"info"` if on monitor mode, empty otherwise
  - scored: `true` if severity set
  - rule: policy name
  - category: either `"validating"` or `"mutating & validating"`

## Test

Tested locally out-of-cluster.

### Potential improvement

~~ClusterPolicyReport doesn't show on policy-reporter.~~ it does.

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -1,6 +1,8 @@
 package report
 
 import (
+	"encoding/json"
+
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
 	"sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1alpha2"
@@ -96,6 +98,22 @@ func (r *PolicyReport) AddResult(policy policiesv1.Policy, resource unstructured
 			Bool("allowed", auditResponse.Response.Allowed).
 			Str("result", string(result.Result)),
 		).Msg("added result to report")
+}
+
+func (r *PolicyReport) GetSummary() (string, error) {
+	marshaled, err := json.Marshal(r.Summary)
+	if err != nil {
+		return "error marshalling summary", err
+	}
+	return string(marshaled), nil
+}
+
+func (r *ClusterPolicyReport) GetSummary() (string, error) {
+	marshaled, err := json.Marshal(r.Summary)
+	if err != nil {
+		return "error marshalling summary", err
+	}
+	return string(marshaled), nil
 }
 
 func (r *ClusterPolicyReport) AddResult(policy policiesv1.Policy, resource unstructured.Unstructured, auditResponse *admv1.AdmissionReview, responseErr error) {
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -129,7 +129,8 @@ func (r *PolicyReport) GetSummaryJSON() (string, error) {
 	return string(marshaled), nil
 }
 
-func (r *ClusterPolicyReport) GetSummary() (string, error) {
+// GetSummaryJSON gets the report.Summary formatted in JSON. Useful for logging
+func (r *ClusterPolicyReport) GetSummaryJSON() (string, error) {
 	marshaled, err := json.Marshal(r.Summary)
 	if err != nil {
 		return "error marshalling summary", err
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -39,7 +39,7 @@ const (
 	SeverityInfo     = "info"
 
 	// Category specifies the category of a policy result
-	CategoryMutating   = "mutating & validating"
+	CategoryMutating   = "mutating"
 	CategoryValidating = "validating"
 
 	LabelAppManagedBy = "app.kubernetes.io/managed-by"
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -23,13 +23,24 @@ type ClusterPolicyReport struct {
 	v1alpha2.ClusterPolicyReport
 }
 
-// Status specifies state of a policy result
 const (
+	// Status specifies state of a policy result
 	StatusPass  = "pass"
 	StatusFail  = "fail"
 	StatusWarn  = "warn"
 	StatusError = "error"
 	StatusSkip  = "skip"
+
+	// Severity specifies severity of a policy result
+	SeverityCritical = "critical"
+	SeverityHigh     = "high"
+	SeverityMedium   = "medium"
+	SeverityLow      = "low"
+	SeverityInfo     = "info"
+
+	// Category specifies the category of a policy result
+	CategoryMutating   = "mutating & validating"
+	CategoryValidating = "validating"
 )
 
 func NewClusterPolicyReport(name string) ClusterPolicyReport {
@@ -116,7 +127,10 @@ func (r *ClusterPolicyReport) GetSummary() (string, error) {
 	return string(marshaled), nil
 }
 
-func (r *ClusterPolicyReport) AddResult(policy policiesv1.Policy, resource unstructured.Unstructured, auditResponse *admv1.AdmissionReview, responseErr error) {
+func (r *ClusterPolicyReport) AddResult(
+	policy policiesv1.Policy, resource unstructured.Unstructured,
+	auditResponse *admv1.AdmissionReview, responseErr error,
+) {
 	result := newPolicyReportResult(policy, resource, auditResponse, responseErr)
 	switch result.Result {
 	case StatusFail:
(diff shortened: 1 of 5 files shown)


internal/report/report.go
```diff
@@ -166,13 +212,19 @@ func newPolicyReportResult(policy policiesv1.Policy, resource unstructured.Unstr
 	}
 
 	return &v1alpha2.PolicyReportResult{
-		Source: PolicyReportSource,
-		Policy: name, // either cap-policy_name or ap-policy_name
+		Source:   PolicyReportSource,
+		Policy:   name,     // either cap-policy_name or ap-policy_name
+		Rule:     rule,     // policy name
```
viccuad, line 217: I'm not entirely happy with using `policy.GetName()` here (notice that this is different from `policy.GetUniqueName()`, that prepends `cap-` or `ap-` for clusterwide or namespaced policies).

I would prefer to take the name of the Wasm module in the registry. Maybe including even the tag. We could use "net/url" and parse the Url for that part, but that's fragile. The best approach would be to be OCI url aware. Maybe using https://github.com/google/go-containerregistry, but that's a big dep just for this. Worth it doing by hand?

internal/report/report.go
```diff
@@ -21,21 +23,38 @@ type ClusterPolicyReport struct {
...
+
+	// Severity specifies severity of a policy result
+	SeverityCritical = "critical"
+	SeverityHigh     = "high"
+	SeverityMedium   = "medium"
+	SeverityLow      = "low"
+	SeverityInfo     = "info"
+
+	// Category specifies the category of a policy result
+	CategoryMutating   = "mutating & validating"
```
viccuad, line 42: bit of a mouthful. Better `mutating then validating`, `mutating+validating`, or just `mutating`?

internal/report/report.go
```diff
@@ -148,14 +192,29 @@ func newPolicyReportResult(policy policiesv1.Policy, resource unstructured.Unstr
...
 	timestamp := *time.ProtoTime()
 
+	var severity v1alpha2.PolicyResultSeverity
+	var scored bool
+	if policy.GetPolicyMode() == policiesv1.PolicyMode(policiesv1.PolicyModeStatusMonitor) {
+		scored = true
+		severity = SeverityInfo
+	}
+
+	var category string
```
viccuad, line 209: Category sadly is a string. It would be nice to have several overlapping categories: mutating, validating, context-aware, and so on.

policy-reporter UI allows filtering by them:

categories

internal/report/report.go
viccuad: After seeing it on policy-reporter filtering (see https://github.com/kubewarden/audit-scanner/pull/47#discussion_r1221453801), I think it should be `mutating`.

flavio: yes, `mutating` should be enough

internal/report/report.go
```diff
@@ -98,7 +120,26 @@ func (r *PolicyReport) AddResult(policy policiesv1.Policy, resource unstructured
 		).Msg("added result to report")
 }
 
-func (r *ClusterPolicyReport) AddResult(policy policiesv1.Policy, resource unstructured.Unstructured, auditResponse *admv1.AdmissionReview, responseErr error) {
+func (r *PolicyReport) GetSummary() (string, error) {
+	marshaled, err := json.Marshal(r.Summary)
```
flavio, line 124: Just out of curiosity, why do we need to marshal that to JSON? Reading the method name I would not expect to get back a JSON encoded string.

I think we should write some inline docs explaining that and, eventually, rename the method to be something like `GetSummaryJSON` (although I'm not super happy/convinced about this name)

internal/report/report.go
```diff
@@ -148,14 +192,29 @@ func newPolicyReportResult(policy policiesv1.Policy, resource unstructured.Unstr
...
 
 	time := metav1.Now()
 	timestamp := *time.ProtoTime()
 
+	var severity v1alpha2.PolicyResultSeverity
+	var scored bool
+	if policy.GetPolicyMode() == policiesv1.PolicyMode(policiesv1.PolicyModeStatusMonitor) {
+		scored = true
+		severity = SeverityInfo
+	}
```
flavio, line 207: Looking into kyverno, this information seems to be taken from an annotation of the policy. For example, looking at [this](https://kyverno.io/policies/other/restrict-secrets-by-label/restrict-secrets-by-label/) policy, you can see the following interesting annotations being specified by the admin:

```yaml
  annotations:
    # We could generate that via kwctl scaffold, using the metadata of the policy 
    policies.kyverno.io/title: Restrict Secrets by Label
    # The meaning of this field is not obvious to me right now, we should document ourselves
    policies.kyverno.io/category: Other
    # This should be set by the admin manually
    policies.kyverno.io/severity: medium
    # AFAIK this is shown by the policy report web UI, isn't it? I think we have that defined inside of the artifacthub metadata, don't we?
    policies.kyverno.io/subject: Pod, Secret
    # We have that, but I would not add it to the annotations of a policy, this would just clutter etcd IMHO
    policies.kyverno.io/description: >-
      Secrets often contain sensitive information and their access should be carefully controlled.
      Although Kubernetes RBAC can be effective at restricting them in several ways,
      it lacks the ability to use labels on referenced entities. This policy ensures
      that only Secrets not labeled with `status=protected` can be consumed by Pods. 
```

To come back to this PR, I would change this code to read the severity from the policy annotation with key `policies.kubewarden.io/severity` (or something like that)

internal/report/report.go
flavio: See the message from above, let's try to understand what is the purpose/logic behind `policies.kyverno.io/category` annotation. Then we can just read it from the policy annotation

internal/report/report.go
flavio: Mmhhh, is kyverno using `policies.kyverno.io/title` maybe?

flavio (commented):
Overall ok. I left some comments. I would expect some small research to be done to better identify the usage of annotations inside of Kyverno's policy definitions. Once we have a better picture we can decide how to move forward.

internal/report/report.go
viccuad: changed to `mutating` in https://github.com/kubewarden/audit-scanner/pull/47/commits/d39bafd678a5f99b70618dae6842dcee8a8617e0.

internal/report/report.go
viccuad: This function is only used (so far) in a log.Debug(), which itself is a json output:
```
{"level":"info","dict":{"report name":"polr-ns-default","report ns":"default","summary":"{\"pass\":0,\"fail\":0,\"warn\":0,\"error\":0,\"skip\":0}"},"time":"2023-06-07T17:57:25+02:00","message":"updated PolicyReport"}
```

I have renamed it and added godocs in https://github.com/kubewarden/audit-scanner/pull/47/commits/09c651a5726c3a76fd0f184bdb90fbef3151f7c2.

internal/report/report.go
viccuad: Agree, sounds better. I suppose it's ok to schedule that after this PR, as we need to change the CRDs definitions on kubewarden-controller.

internal/report/report.go
viccuad: I think you are right. I would leave it as it is in the PR.

internal/report/report.go
```diff
@@ -98,7 +120,27 @@ func (r *PolicyReport) AddResult(policy policiesv1.Policy, resource unstructured
...
+// GetSummaryJSON gets the report.Summary formatted in JSON. Useful for logging
+func (r *PolicyReport) GetSummaryJSON() (string, error) {
+	marshaled, err := json.Marshal(r.Summary)
+	if err != nil {
+		return "error marshalling summary", err
+	}
+	return string(marshaled), nil
+}
+
+func (r *ClusterPolicyReport) GetSummary() (string, error) {
```
jvanz, line 132: Shouldn't we rename this to `GetSummaryJSON` as well?

viccuad:
WRT to the ClusterPolicyReports showing on policy-reporter, non-empty ones show fine.

internal/report/report.go
viccuad: indeed! done in https://github.com/kubewarden/audit-scanner/pull/47/commits/8421778c7678549954128966d91d3308a28a7422.

internal/report/report.go
viccuad: >  AFAIK this is shown by the policy report web UI, isn't it? I think we have that defined inside of the artifacthub metadata, don't we?
>    policies.kyverno.io/subject: Pod, Secret

Yes. I think this doesn't need additional work. For the PolicyReportResult subject, it is generated with the ObjectReference i has, which I think is a better fit.


For the rest, I have opened https://github.com/kubewarden/kubewarden-controller/issues/461 and added it to the epic, please have a look.

jvanz (approved):

flavio (approved):

PR #48, opened 2023-06-08
ClusterWide resources scan.

### LINKED ISSUE #43: audit-scanner - MVP Stage 5.1 - inspect cluster wide resources
Evolve the audit-scanner to be able to inspect cluster-wide resources.
Note: this can be done in parallel of stage 5.2
    
### Acceptance criteria:
- [x] The cli allows to either scan a specific namespace, or inspect cluster-wide resources. This depends on how the program is invoked by the user
- [x] When scanning cluster-wide resources, a `ClusterPolicyReport` is generated
- [x] We have unit tests
- [x] We have integration test **if needed**

jvanz:
## Description

Add a new option in the audit scanner to scan only cluster wide resources.

Fix #43


jvanz:
I'm aware that the unit tests are failing. That's weird because the tests are passing locally and no error is shown in the CI execution log. When I've added a debug logs in the test I found out that the execution stopped on around this [line](https://github.com/kubewarden/audit-scanner/blob/f3a59d33daf62e7d614443da65907cacbfa453d6/internal/policies/fetcher_test.go#L333). I don't know what is happening yet. But I'll continue to investigate

jvanz:
@kubewarden/kubewarden-developers  I'm aware that the unit tests are failing. That's weird because the tests are passing locally and no error is shown in the CI execution log. When I've added a debug logs in the test I found out that the execution stopped on around this [line](https://github.com/kubewarden/audit-scanner/blob/f3a59d33daf62e7d614443da65907cacbfa453d6/internal/policies/fetcher_test.go#L333). I don't know what is happening yet. But I'll continue to investigate

cmd/root_test.go
```diff
@@ -50,3 +53,7 @@ func (s *mockScanner) ScanAllNamespaces() error {
 	s.scanAllNamespacesCalled = true
 	return nil
 }
+
+func (s *mockScanner) ScanClusterWideResources() error {
+	return fmt.Errorf("mockScanner has not implemented ScanClusterWideResources yet")
```
flavio, line 58: Shouldn't this be done as part of this PR?

internal/errors/errors.go
```diff
@@ -0,0 +1,5 @@
+package errors
+
+import "errors"
+
+var ErrResourceNotFound = errors.New("resource not found")
```
flavio, line 5: While this isn't actually a constant, it's being used like one. Maybe we should just move it to the `constants.go` file. It's also a bit weird (but that's a Go problem), to have a file named `error.go` that imports a package named in the same way

internal/policies/fetcher.go
```diff
@@ -63,6 +63,30 @@ func (f *Fetcher) GetPoliciesForANamespace(namespace string) ([]policiesv1.Polic
 	return filteredPolicies, skippedNum, nil
 }
 
+func (f *Fetcher) getClusterAdmissionPolicies() ([]policiesv1.ClusterAdmissionPolicy, error) {
+	policies := &policiesv1.ClusterAdmissionPolicyList{}
+	err := f.client.List(context.Background(), policies)
+	if err != nil {
+		return nil, err
```
flavio, line 70: We should not return `nil`, but an empty list

internal/policies/fetcher.go
```diff
@@ -63,6 +63,30 @@ func (f *Fetcher) GetPoliciesForANamespace(namespace string) ([]policiesv1.Polic
...
+	}
+	return policies.Items, nil
+}
+
+func (f *Fetcher) GetClusterAdmissionPolicies() ([]policiesv1.Policy, int, error) {
+	clusterAdmissionPolicies, err := f.getClusterAdmissionPolicies()
+	if err != nil {
+		return nil, 0, err
```
flavio, line 78: We should not return `nil`, but an empty list

internal/policies/fetcher.go
```diff
@@ -63,6 +63,30 @@ func (f *Fetcher) GetPoliciesForANamespace(namespace string) ([]policiesv1.Polic
...
+func (f *Fetcher) getClusterAdmissionPolicies() ([]policiesv1.ClusterAdmissionPolicy, error) {
+	policies := &policiesv1.ClusterAdmissionPolicyList{}
+	err := f.client.List(context.Background(), policies)
+	if err != nil {
+		return nil, err
+	}
+	return policies.Items, nil
+}
+
+func (f *Fetcher) GetClusterAdmissionPolicies() ([]policiesv1.Policy, int, error) {
```
flavio, line 75: please add some inline documentation explaining what the function is doing and what is returning

internal/policies/fetcher.go
```diff
@@ -63,6 +63,30 @@ func (f *Fetcher) GetPoliciesForANamespace(namespace string) ([]policiesv1.Polic
...
+	clusterAdmissionPolicies, err := f.getClusterAdmissionPolicies()
+	if err != nil {
+		return nil, 0, err
+	}
+	policies := []policiesv1.Policy{}
+	for _, policy := range clusterAdmissionPolicies {
+		policy := policy
+		policies = append(policies, &policy)
+	}
+	filteredPolicies := f.filter(policies)
```
flavio, line 85: Unrelated with this PR. I was having a hard time figuring out why we should be filtering the cluster policies. In the end it makes sense, but there's a layer of indirection introduced by having `f.filter` be a function pointer.

Do we really need that? Can't we just invoke the `filterAuditablePolicies` function? That would make the code more readable IMHO

internal/resources/fetcher.go
```diff
@@ -2,18 +2,22 @@ package resources
...
 	"fmt"
 	"net/url"
 
+	"github.com/kubewarden/audit-scanner/internal/constants"
+	auditscannererrors "github.com/kubewarden/audit-scanner/internal/errors"
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	"github.com/rs/zerolog/log"
 	v1 "k8s.io/api/core/v1"
-	"k8s.io/apimachinery/pkg/api/errors"
+	apimachineryerrors "k8s.io/apimachinery/pkg/api/errors"
```
flavio, line 14: Reminder: if we get rid of the `errors.go` file from this codebase, this aliasing is not required

internal/resources/fetcher.go
```diff
@@ -58,12 +64,46 @@ func NewFetcher(kubewardenNamespace string, policyServerURL string) (*Fetcher, e
...
+	}
+	for _, apiResource := range apiResourceList.APIResources {
+		if apiResource.Name == gvr.Resource {
+			return apiResource.Namespaced, nil
+		}
+	}
+	return false, auditscannererrors.ErrResourceNotFound
+}
+
+// Method used by other method to fetch resources evaluated by the given policies.
```
flavio, line 86: nitpicking:

```suggestion
// Given a list of policies and a namespace, return the list of resources that have to be audited
```

internal/resources/fetcher.go
```diff
@@ -58,12 +64,46 @@ func NewFetcher(kubewardenNamespace string, policyServerURL string) (*Fetcher, e
...
+				if errors.Is(err, auditscannererrors.ErrResourceNotFound) {
+					log.Warn().Msg(fmt.Sprintf("API resource (%s) not found", gvr.String()))
+					continue
+				}
+				return nil, err
+			}
+			if isNamespaced {
+				continue
+			}
+		}
```
flavio, line 103: I think this deserves to be placed into a dedicated method, like `getClusterwideResources`, with an explanation about why this is different from `getResources`

internal/resources/fetcher.go
```diff
@@ -80,6 +120,10 @@ func (f *Fetcher) GetResourcesForPolicies(ctx context.Context, policies []polici
 	return auditableResources, nil
 }
 
+func (f *Fetcher) GetClusterWideResourcesForPolicies(ctx context.Context, policies []policiesv1.Policy) ([]AuditableResources, error) {
+	return f.getResources(ctx, policies, constants.NoNamespace)
```
flavio, line 124: Thinking about what I've commented above, this is where I would move all the code

flavio (changes requested):

internal/policies/fetcher.go
viccuad: in the future, we could use the filter function to cribe for policies, in addition or regardless of their `spec.ContextAwareResources`. For example we could try to run all policies against all resources: if the resource is listed in `spec.ContextAwareResources`, we evaluate as usual. If it is not, we do our best, and if we get an AdmissionReview with an error back, we ignore the result. This would allow us to naively audit as much things as possible.

But agree, for now, I can't find a utility and it doesn't make the code more readable.

internal/resources/fetcher.go
```diff
@@ -88,8 +132,14 @@ func (f *Fetcher) getResourcesDynamically(ctx context.Context,
 		Version:  version,
 		Resource: resource,
 	}
-	list, err := f.dynamicClient.Resource(resourceID).Namespace(namespace).
-		List(ctx, metav1.ListOptions{})
+	var list *unstructured.UnstructuredList
+	var err error
+	// if no namespace, we need to look for cluster wide resources
+	if len(namespace) != 0 {
```
viccuad, line 138: I would  prefer `getResourcesDynamically(ns) and `getClusterwideResourcesDynamically()`.

internal/scanner/scanner.go
```diff
@@ -109,6 +113,67 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+				}
+			}
+		}
+	}
+	err = s.reportStore.AddClusterPolicyReport(&clusterReport)
+	if err != nil {
+		log.Error().Err(err).Msg("error adding PolicyReport to store")
+	}
+	// TODO for debug
+	str, err := s.reportStore.ToJSON()
```
viccuad, line 168: I would do this only when `s.printJSON` is true, as in ScanNamespace()

internal/scanner/scanner.go
```diff
@@ -109,6 +113,67 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+					log.Debug().Dict("response", zerolog.Dict().
+						Str("uid", string(auditResponse.Response.UID)).Bool("allowed", auditResponse.Response.Allowed).Str("policy", policy.GetName()).Str("resource", resource.GetName()),
+					).
+						Msg("audit review response")
+					clusterReport.AddResult(policy, resource, auditResponse, responseErr)
+				}
+			}
+		}
+	}
+	err = s.reportStore.AddClusterPolicyReport(&clusterReport)
```
viccuad, line 163: maybe this should be `s.reportStore.UpdateClusterPolicyReport()`. Right now both `AddClusterPolicyReport()` and `UpdateClusterPolicyReport()` are the same, but that doesn't mean it is the case in the future. Also, if the ClusterPolicyReport already exists because of a prior scan, we should be updating it.

internal/scanner/scanner.go
```diff
@@ -109,6 +113,67 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+			if err != nil {
+				// TODO what's the better thing to do here?
+				log.Error().Err(err)
+				continue
+			}
+			for _, resource := range resource.Resources {
+				admissionRequest := resources.GenerateAdmissionReview(resource)
+				auditResponse, responseErr := sendAdmissionReviewToPolicyServer(url, admissionRequest, &s.httpClient)
+				if responseErr != nil {
+					// log error, will end in PolicyReportResult too
```
viccuad, line 148:
```suggestion
					// log error, will end in ClusterPolicyReportResult too
```

internal/scanner/scanner.go
```diff
@@ -109,6 +113,67 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+	if err != nil {
+		log.Error().Err(err).Msg("error adding PolicyReport to store")
+	}
+	// TODO for debug
+	str, err := s.reportStore.ToJSON()
+	fmt.Println(str)
+	if err != nil {
+		log.Error().Err(err).Msg("error marshaling reportStore to JSON")
+	}
+	log.Info().Msg("scan finished")
```
viccuad, line 173: I would move the log.Info() just right before s.reportStore.AddClusterPolicyReport() to match how it is in ScanNamespace() (perform scan, save scan to policyreports). My rationale is that reportStore functions have their own logs, and in the future they may be more convoluted (sql backend, etc).

internal/scanner/scanner.go
```diff
@@ -109,6 +113,67 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+	if err != nil {
+		return err
+	}
+	// create PolicyReport
+	clusterReport := report.NewClusterPolicyReport("cluster-report")
+	clusterReport.Summary.Skip = skippedNum
+	// Iterate through all auditableResources. Each item contains a list of resources and the policies that would need
+	// to evaluate them.
+	for _, resource := range auditableResources {
+		for _, policy := range resource.Policies {
```
viccuad, line 137: Optional nitpicking, feel free to ignore: I would create a new function for this loop, analogous to `auditResource()`, (e.g: clusterAuditResource() or `auditClusterResource()`). This would reduce the complexity of `ScanClusterWideResources()`, and separate the "map" and "reduce" logic.

internal/resources/fetcher.go
jvanz: Not really. I'm using the `errors.Is` function in the code. Thus, we still need the alias.

jvanz:
@kubewarden/kubewarden-developers the issue with the unit tests running on CI was that the function used to create the fetcher call a k8s lib which panics if the kube config file does not exists. Therefore, the sub tests were failing with no error message.

internal/policies/fetcher.go
flavio: Let's leave the code as it is in the meantime. The complexity is not so big

internal/resources/fetcher.go
```diff
@@ -88,8 +151,29 @@ func (f *Fetcher) getResourcesDynamically(ctx context.Context,
 		Version:  version,
 		Resource: resource,
 	}
-	list, err := f.dynamicClient.Resource(resourceID).Namespace(namespace).
-		List(ctx, metav1.ListOptions{})
+	var list *unstructured.UnstructuredList
+	var err error
+	// if no namespace, we need to look for cluster wide resources
```
flavio, line 156: I think this comment is a leftover from the initial PR. I think it can be removed now

internal/resources/fetcher_test.go
```diff
@@ -284,3 +288,203 @@ func TestGetServiceByAppLabel(t *testing.T) {
...
+					Kind:         "Namespace",
+					Namespaced:   false,
+				},
+				{
+					Name:         "pods",
+					SingularName: "pod",
+					Kind:         "Pod",
+					Namespaced:   true,
+				},
+			},
```
flavio, line 462: Shouldn't this be an empty list?

internal/scanner/scanner.go
viccuad: Yep, I started the same interface on a previous PR, but we removed it at the end. I'm ok adding it later on.

internal/resources/fetcher.go
```diff
@@ -29,6 +32,7 @@ type Fetcher struct {
 	// FQDN of the policy server to query. If not empty, Fetcher will query on
 	// port 3000. Useful for out-of-cluster debugging
 	policyServerURL string
+	clientset       kubernetes.Interface
```
viccuad, line 35: nit: Maybe worth documenting the struct fields on why we have the dynamic client (for CRDs) and the normal client (for discovery).

viccuad (approved):
LGTM, modulo the current comments!

internal/resources/fetcher_test.go
jvanz: This is the  list returned by the k8s api mock. Thus, the audit scanner should validate that resources returned by the API contains or not the resource requested. You can notice that in the `expectedIsNamespaced` and `expectedErr` fields. Where the expected values are `false` and `constants.ErrResourceNotFound` respectively.

PR #51, opened 2023-06-14
feat: reuse report results.

### LINKED ISSUE #44: audit-scanner - MVP Stage 5.2 - reuse results from previous evaluations
Leverage previous `PolicyReport` objects as a cache for evaluations
Note: this can be done in parallel of stage 5.1

### Acceptance criteria:
 - See [step 5 of audit RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0011-audit-checks.md#step-5-attempt-to-reuse-previous-evaluation-results)
 - We have unit tests
 - We have integration test **if needed**

jvanz:
## Description

The audit scanner now check if there is a policy evaluation result in the previous scan report for the same policy and resource version tuple. If there is, the result is copy into the new report.

Fix #44

### DIFF
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -2,9 +2,11 @@ package report
 
 import (
 	"encoding/json"
+	"fmt"
 
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/types"
 	"sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1alpha2"
 
 	"github.com/kubewarden/audit-scanner/internal/constants"
@@ -54,7 +56,7 @@ func NewClusterPolicyReport(name string) ClusterPolicyReport {
 	return ClusterPolicyReport{
 		ClusterPolicyReport: v1alpha2.ClusterPolicyReport{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:              PrefixNameClusterPolicyReport + name,
+				Name:              getClusterReportName(name),
 				CreationTimestamp: metav1.Now(),
 				Labels:            labels,
 			},
@@ -76,7 +78,7 @@ func NewPolicyReport(namespace *v1.Namespace) PolicyReport {
 	return PolicyReport{
 		PolicyReport: v1alpha2.PolicyReport{
 			ObjectMeta: metav1.ObjectMeta{
-				Name:              PrefixNamePolicyReport + namespace.Name,
+				Name:              getNamespacedReportName(namespace.Name),
 				Namespace:         namespace.Name,
 				CreationTimestamp: metav1.Now(),
 				Labels:            labels,
@@ -101,6 +103,26 @@ func NewPolicyReport(namespace *v1.Namespace) PolicyReport {
 	}
 }
 
+func getClusterReportName(name string) string {
+	return PrefixNameClusterPolicyReport + name
+}
+
+func getNamespacedReportName(namespace string) string {
+	return PrefixNamePolicyReport + namespace
+}
+
+func (r *PolicyReport) AddResult(result *v1alpha2.PolicyReportResult) {
+	switch result.Result {
+	case StatusFail:
+		r.Summary.Fail++
+	case StatusError:
+		r.Summary.Error++
+	case StatusPass:
+		r.Summary.Pass++
+	}
+	r.Results = append(r.Results, result)
+}
+
 // GetSummaryJSON gets the report.Summary formatted in JSON. Useful for logging
 func (r *PolicyReport) GetSummaryJSON() (string, error) {
 	marshaled, err := json.Marshal(r.Summary)
@@ -119,29 +141,24 @@ func (r *ClusterPolicyReport) GetSummaryJSON() (string, error) {
 	return string(marshaled), nil
 }
 
-func (r *PolicyReport) AddResult( //nolint:dupl
+// GetReusablePolicyReportResult tries to find a PolicyReportResult that
+// can be reused.
+//
+// The result can be reused if both these conditions are
+// satisfied:
+//   - The subject of the PolicyReportResult (the object that was inspected)
+//     has not been changed since the report was created
+//   - The policy that evaluated the subject (now given by the user as
+//     parameter) has not been changed since the report was created
+func (r *PolicyReport) GetReusablePolicyReportResult(policy policiesv1.Policy, resource unstructured.Unstructured) *v1alpha2.PolicyReportResult {
+	return findReusableResult(r.Results, policy, resource)
+}
+
+func (r *PolicyReport) CreateResult(
 	policy policiesv1.Policy, resource unstructured.Unstructured,
 	auditResponse *admv1.AdmissionReview, responseErr error,
-) {
+) *v1alpha2.PolicyReportResult {
 	result := newPolicyReportResult(policy, resource, auditResponse, responseErr)
-	switch result.Result {
-	case StatusFail:
-		r.Summary.Fail++
-	case StatusError:
-		r.Summary.Error++
-	case StatusPass:
-		r.Summary.Pass++
-	default:
-		// this should never happen
-		log.Error().
-			Str("report name", r.Name).
-			Dict("result", zerolog.Dict().
-				Str("policy", policy.GetName()).
-				Str("resource", resource.GetName()).
-				Str("result", string(result.Result)),
-			).Msg("result unknown")
-	}
-	r.Results = append(r.Results, result)
 	log.Debug().
 		Str("report name", r.Name).
 		Dict("result", zerolog.Dict().
@@ -150,13 +167,23 @@ func (r *PolicyReport) AddResult( //nolint:dupl
 			Bool("allowed", auditResponse.Response.Allowed).
 			Str("result", string(result.Result)),
 		).Msg("added result to report")
+	return result
 }
 
-func (r *ClusterPolicyReport) AddResult( //nolint:dupl
+func (r *ClusterPolicyReport) CreateResult(
 	policy policiesv1.Policy, resource unstructured.Unstructured,
 	auditResponse *admv1.AdmissionReview, responseErr error,
-) {
+) *v1alpha2.PolicyReportResult {
 	result := newPolicyReportResult(policy, resource, auditResponse, responseErr)
+	log.Debug().Str("report name", r.Name).Dict("result", zerolog.Dict().
+		Str("policy", policy.GetName()).Str("resource", resource.GetName()).
+		Bool("allowed", auditResponse.Response.Allowed).
+		Str("result", string(result.Result)),
+	).Msg("added result to report")
+	return result
+}
+
+func (r *ClusterPolicyReport) AddResult(result *v1alpha2.PolicyReportResult) {
 	switch result.Result {
 	case StatusFail:
 		r.Summary.Fail++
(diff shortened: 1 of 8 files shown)


internal/report/report_test.go
```diff
@@ -51,3 +56,252 @@ func TestAddResult(t *testing.T) {
...
+			"status": map[string]interface{}{},
+		},
+		}
+		report.AddResult(&policy, namespaceResource, &admReviewPass, nil)
+		expectedResource = namespaceResource
+	}
+
+	result, ok := report.GetResult(&policy, expectedResource)
+	if !ok {
+		t.Errorf("Policy result not found")
```
viccuad, line 113: We should add needed fields to all the logs: policy & resource for result not found, report name for report not found, etc (this applies to the rest of log calls).

internal/scanner/scanner.go
```diff
@@ -89,11 +89,16 @@ func (s *Scanner) ScanNamespace(nsName string) error {
 	// create PolicyReport
 	namespacedsReport := report.NewPolicyReport(namespace)
 	namespacedsReport.Summary.Skip = skippedNum
+	// old policy report to be used as cache
+	previousNamespacedReport, err := s.reportStore.GetPolicyReport(nsName)
+	if err != nil {
+		log.Error().Err(err).Msg("error getting PolicyReport from store")
```
viccuad, line 95: I'm getting always a log error here (I believe). We just add fields to the log, in this case the report name, or the namespace.

viccuad, line 95: Testing it locally, with an existing PolicyReport, and it always errors here, hence it never uses the previous report.

internal/report/report.go
```diff
@@ -237,9 +296,12 @@ func newPolicyReportResult( //nolint:funlen
...
 		Scored:          scored,
-		Subjects:        []*v1.ObjectReference{resourceObjectReference}, // reference to object evaluated
+		Subjects:        []*v1.ObjectReference{&resourceObjectReference}, // reference to object evaluated
 		SubjectSelector: &metav1.LabelSelector{},
 		Description:     description, // output message of the policy
-		Properties:      map[string]string{},
+		// The policy resource version is used to check if the same result can used
+		// in the next scan
+		// https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency
+		Properties: map[string]string{PropertyPolicyResourceVersion: policy.GetResourceVersion()},
```
viccuad, line 305: I'm not sure this is correct; I don't think we should save policy resource versions on the result.
Properties could be end-user readable.

internal/report/report.go
```diff
@@ -162,7 +179,72 @@ func (r *ClusterPolicyReport) AddResult(
...
+		Kind:    constants.KubewardenKindClusterAdmissionPolicy,
+	}:
+		return "cap-" + policy.GetName(), nil
+	case schema.GroupVersionKind{
+		Group:   constants.KubewardenPoliciesGroup,
+		Version: constants.KubewardenPoliciesVersion,
+		Kind:    constants.KubewardenKindAdmissionPolicy,
+	}:
+		return "ap-" + policy.GetName(), nil
+	default:
```
viccuad, line 230: I would add back the `// this never happens` comment.

internal/scanner/scanner.go
```diff
@@ -151,26 +161,46 @@ func (s *Scanner) ScanClusterWideResources() error {
...
 			// TODO what's the better thing to do here?
 			log.Error().Err(err)
 			continue
 		}
 		for _, resource := range resource.Resources {
+			if result, foundResult := previousClusterReport.GetResult(policy, resource); foundResult {
+				policyResourceVersion, ok := result.Properties[report.PropertyPolicyResourceVersion]
+				if ok && policyResourceVersion == policy.GetResourceVersion() {
+					// We have a result from the same policy version for the same resource instance.
+					// Skip the evaluation
```
viccuad, line 177: I think this isn't correct. We are looking at the policy version, but the resources may have changed, hence the PolicyReportResult for that policy may change too.

According to the RFC, https://github.com/kubewarden/rfc/blob/main/rfc/0011-audit-checks.md#step-5-attempt-to-reuse-previous-evaluation-results:

> If there's an entry, look at the evaluation timestamp of this entry. If policy update timestamp and pod instance update timestamp are older than the evaluation timestamp it means that nothing changed.

I think we need to look at the PolicyReportResult.timestamp, and use that to compare against the policy update timestamp and resource update timestamp, which have nothing to do with the resource version (generation).

viccuad (changes requested):

internal/scanner/scanner.go
jvanz: I'm aware that this change is not described in the RFC. However, I've spent some time trying to figuring out what would be the best "timestamp" field to be used. Let's use a [deployment](https://pkg.go.dev/k8s.io/api/apps/v1#Deployment) kind as example. There is just one timestamp, the [CreationTimestamp](https://pkg.go.dev/k8s.io/apimachinery/pkg/apis/meta/v1#ObjectMeta). But this field has only the time when the resource is created. Not when it's updated. Thus, in the approached described in the RFC we could ended with  always reusing the old result. Even if the user changed some field in the deployment object. The same apply for the policy resource. 

The only field that I found that is updated on any change in the resourceVersion. See more in this [doc](https://github.com/kubernetes/community/blob/master/contributors/devel/sig-architecture/api-conventions.md#concurrency-control-and-consistency):

> All Kubernetes resources have a "resourceVersion" field as part of their metadata. This resourceVersion is a string that identifies the internal version of an object that can be used by clients to determine when objects have changed

> The resourceVersion is changed by the server every time an object is modified. 

This makes sense if you consider the [ObjectReference](https://pkg.go.dev/k8s.io/api@v0.27.3/core/v1#ObjectReference) type. There is a field for resource version there. And the policy result type uses this references in the `Subjects` field. Thus, I believe that the use of the `ResourceVersion` is not incorrect. 

Am I missing something? If so, what is the timestamp field that you have in mind to be used in the comparison?

internal/scanner/scanner.go
```diff
@@ -151,26 +162,49 @@ func (s *Scanner) ScanClusterWideResources() error {
...
 			// TODO what's the better thing to do here?
 			log.Error().Err(err)
 			continue
 		}
 		for _, resource := range resource.Resources {
+			if previousClusterReport != nil {
+				if result, foundResult := previousClusterReport.GetResult(policy, resource); foundResult {
+					fmt.Println(result)
+					policyResourceVersion, ok := result.Properties[report.PropertyPolicyResourceVersion]
+					if ok && policyResourceVersion == policy.GetResourceVersion() {
```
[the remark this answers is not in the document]
jvanz: This piece of code does not check the resource `resourceVersion`. But the function `findResult` called from `GetResult`  compares the resource object reference which uses the `resourceVersion` field.

internal/scanner/scanner.go
jvanz: JFYI: https://github.com/kubewarden/audit-scanner/pull/51#discussion_r1231090070

viccuad: I see. For a moment I though about the Status subresource, but that may not be present. Maybe it is worth to have a look on how others have implemented it.

jvanz: Let me check if I can find something... another option is to have a maximum time window witch the scanner can reuse result. For example, if the result is 1day old, reevaluate. This time window can be configurable by the user.

internal/report/store.go
```diff
@@ -89,23 +89,62 @@ func (s *PolicyReportStore) AddClusterPolicyReport(report *ClusterPolicyReport)
 	return nil
 }
 
+func (s *PolicyReportStore) lookForNamespacedReport(namespace string) (PolicyReport, bool) {
+	s.namespacedPolicyReportsMutex.Lock()
+	defer s.namespacedPolicyReportsMutex.Unlock()
```
flavio, line 94: This is a read operation, you should take a read lock `.RLock()`

internal/report/store.go
```diff
@@ -89,23 +89,62 @@ func (s *PolicyReportStore) AddClusterPolicyReport(report *ClusterPolicyReport)
...
-	s.namespacedPolicyReportsMutex.RLock()
-	defer s.namespacedPolicyReportsMutex.RUnlock()
-	report, present := s.namespacedPolicyReports[namespace]
-	if present {
+	// check if we already have a report to the given namespace in the cache
+	if report, ok := s.lookForNamespacedReport(namespace); ok {
 		return report, nil
 	}
-	return PolicyReport{}, errors.New("report not found")
+	// nothing in the cache. Let's request it in the cluster
```
flavio, line 107: The Kubernetes Go client already keeps its own cache, and uses watches to keep it up to date. Do we need to have our own cache for the PolicyReports? I would always try to get the PolicyReport by querying the client. In some cases this is going to use its own cache, when needed it will will hit the API server.

internal/report/store.go
```diff
@@ -89,23 +89,62 @@ func (s *PolicyReportStore) AddClusterPolicyReport(report *ClusterPolicyReport)
...
+	return *policyReport, nil
 }
 
 // Get the ClusterPolicyReport
 func (s *PolicyReportStore) GetClusterPolicyReport() (ClusterPolicyReport, error) {
 	s.clusterPolicyReportMutex.RLock()
 	defer s.clusterPolicyReportMutex.RUnlock()
-	report := s.clusterPolicyReport
-	return report, nil
+	fmt.Print(s.clusterPolicyReport.ClusterPolicyReport.ObjectMeta.Name)
```
flavio, line 133: leftover, or use the logger with debug level

internal/report/store.go
```diff
@@ -89,23 +89,62 @@ func (s *PolicyReportStore) AddClusterPolicyReport(report *ClusterPolicyReport)
...
+	if len(s.clusterPolicyReport.ClusterPolicyReport.Results) == 0 {
```
flavio, line 134: Also in this case, I would just use the cache provided by the kubernetes' go client

internal/scanner/scanner.go
flavio: I too think we have to use the `ResourceVersion` to understand if a resource has changed since the last time we looked at it

internal/report/report.go
flavio: I think we should use the `ResourceRevision` (see comment below), but we should not store that inside of the `Properties`. Rather, we should use the dedicated field inside of the `ObjectReference` struct: https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference

internal/report/store.go
jvanz: Sorry folks, I agree with you. I don't know why I've kept the old in memory map . In my mind I though that we would like to keep that for some reason that I cannot remember. Any way, as we are using etcd as the backend, it does not make sense to have it if we can send data directly to the backend.

jvanz:
> I think we should use the ResourceRevision (see comment below), but we should not store that inside of the Properties. Rather, we should use the dedicated field inside of the ObjectReference struct: https://pkg.go.dev/k8s.io/api/core/v1#ObjectReference

@flavio, I gree that an `ObjectReference` to the policy is better. But where to store it? The [PolicyReportResult CRD](https://htmlpreview.github.io/?https://github.com/kubernetes-sigs/wg-policy-prototypes/blob/master/policy-report/docs/index.html#wgpolicyk8s.io/v1beta1.*sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1beta1.PolicyReportResult) has only a policy field name. The `resources` field aims to store references to the checked resources. We can customize the usage of this field and add the `ObjectReference` to the policy. But then we will diverge from the original goal of the field. That's the reason I've suggested to create a proposal to the wg-policy  to add an optional field to keep the policy's `ObjectReference`. 

Another workaround is to store all the `ObjectReference` fields individually in the `properties`  PolicyReportResult CRD field and build the `ObjectReference` object in runtime. Something similar to what we already did for the `resourceVersion`

flavio:
Let's try to recap the situation.

We need to figure out if we can reuse an old evaluation result. The evaluation result has the following data:

* the object being inspected
* the policy that evaluated it
* the timestamp that describes when the evaluation took place

From my understanding, we cannot use the evaluation timestamp because we cannot find a "trusted timestamp" that indicates when neither the policy nor the object being inspected was changed.

Because of that, the PR proposes to use the revision counter of the two resources (the policy and the object being inspected).

Due to the nature of the `PolicyReportResult` CR, we could use the following approach:

* Store the revision of the object being inspected inside of the `(Cluster)PolicyReport.scope.resourceRevision`
* Store the revision of the policy inside of a custom property defined inside of `PolicyReportResult.properties`. I don't see any issue with introducing this new property, afterall the `properties` field is a free-form dictionary

I guess that should be enough, isn't it?

flavio:
As discussed during the daily, let's store the UID of the policy + its revision number inside of the policy report properties. Each one of them is going to be stored inside of its own dedicated property

jvanz:
> As discussed during the daily, let's store the UID of the policy + its revision number inside of the policy report properties. Each one of them is going to be stored inside of its own dedicated property

Done

internal/scanner/scanner.go
```diff
@@ -89,27 +89,24 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+		log.Error().Err(err).Str("namespace", nsName).
+			Msg("error getting PolicyReport from store")
+	}
 
 	// Iterate through all auditableResources. Each item contains a list of resources and the policies that would need
 	// to evaluate them.
 	for i := range auditableResources {
-		auditResource(&auditableResources[i], &s.resourcesFetcher, &s.httpClient, &namespacedsReport)
-		err = s.reportStore.AddPolicyReport(&namespacedsReport)
+		auditResource(&auditableResources[i], s.resourcesFetcher, &s.httpClient, &namespacedsReport, &previousNamespacedReport)
```
flavio, line 102: probably not for this PR, but I would like to refactor the `auditResource` function. I think it should become a method of `AuditableResource`, something like `evaluate` or `process`

flavio (approved):
I think this is good to merge, but I would like @viccuad to review the code too.

I left some minor comments and I pushed a small change that does different refactors of the code. Nothing significant, just some renames and some inline docs

internal/scanner/scanner.go
jvanz: I agree. I've also suggested to do a change in the future adding an abstraction layer to avoid this duplicate function to handle Cluster and not cluster reports. They are almost the same.

flavio:
@jvanz please, don't force push again. You wiped out the commit I pushed with the small refactoring I made. Luckily I still had it around and I was able to apply it without conflicts.

I you want to do more changes do a regular pull on your branch, it should work

internal/scanner/scanner.go
```diff
@@ -128,49 +125,64 @@ func (s *Scanner) ScanClusterWideResources() error {
 		return err
 	}
 	// create PolicyReport
-	clusterReport := report.NewClusterPolicyReport("cluster-report")
+	clusterReport := report.NewClusterPolicyReport("clusterwide")
```
viccuad, line 128: Maybe this should be part of report/constants.go.

internal/scanner/scanner.go
```diff
@@ -89,27 +89,24 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
+			Msg("error getting PolicyReport from store")
+	}
 
 	// Iterate through all auditableResources. Each item contains a list of resources and the policies that would need
 	// to evaluate them.
 	for i := range auditableResources {
-		auditResource(&auditableResources[i], &s.resourcesFetcher, &s.httpClient, &namespacedsReport)
-		err = s.reportStore.AddPolicyReport(&namespacedsReport)
+		auditResource(&auditableResources[i], s.resourcesFetcher, &s.httpClient, &namespacedsReport, &previousNamespacedReport)
+		err = s.reportStore.SavePolicyReport(&namespacedsReport)
```
viccuad, line 103: question: do we want to remove the in memory store?

I thought that the idea was to have the in memory store so one can store each evaluation, and only do a client create or update when needing, blocking less.
I also thought we would keep the reportStore.ToJSON() for debugging and for consuming it by another component if needed.

In my opinion I would keep the in-mem store, and a new client store, with an interface for both.

internal/scanner/scanner.go
```diff
@@ -89,27 +89,24 @@ func (s *Scanner) ScanNamespace(nsName string) error {
...
 		if err != nil {
 			log.Error().Err(err).Msg("error adding PolicyReport to store")
 		}
 	}
 	log.Info().Str("namespace", nsName).Msg("scan finished")
-
-	if s.printJSON {
```
viccuad, line 104: if we aren't using printJSON anymore, then I would remove the cli flag, and the cmd tests too.

But as stated in the previous comment I would prefer to have both the in memory store, and a new client store, that both implement an interface for saving the reports.

internal/scanner/scanner.go
```diff
@@ -128,49 +125,64 @@ func (s *Scanner) ScanClusterWideResources() error {
...
+	}
 	// Iterate through all auditableResources. Each item contains a list of resources and the policies that would need
 	// to evaluate them.
 	for i := range auditableResources {
-		auditClusterResource(&auditableResources[i], &s.resourcesFetcher, &s.httpClient, &clusterReport)
+		auditClusterResource(&auditableResources[i], s.resourcesFetcher, &s.httpClient, &clusterReport, &previousClusterReport)
 	}
 	log.Info().Msg("scan finished")
-	err = s.reportStore.UpdateClusterPolicyReport(&clusterReport)
+	err = s.reportStore.SaveClusterPolicyReport(&clusterReport)
```
viccuad, line 141: same concerns about substituting the in-mem store, instead of providing a new client store.

internal/scanner/scanner_test.go
```diff
@@ -1,43 +1,54 @@
...
 	}
 }
 
+var defaultHTTPTestingServerHandler = http.HandlerFunc(defaultHTTPTestingServerFunc)
+
+type ResourcesFetcherMock struct {
+	auditURL string
+}
+
+func (r ResourcesFetcherMock) GetResourcesForPolicies(_ context.Context, _ []policiesv1.Policy, _ string) ([]resources.AuditableResources, error) {
```
viccuad, line 37: I think that by having an in-mem store, that can be better mocked, we may not need this mocks.

internal/scanner/scanner_test.go
```diff
@@ -1,43 +1,54 @@
...
+	return url.Parse(r.auditURL)
+}
+
+func (r ResourcesFetcherMock) GetClusterWideResourcesForPolicies(_ context.Context, _ []policiesv1.Policy) ([]resources.AuditableResources, error) {
+	return []resources.AuditableResources{}, nil
+}
+
 func TestSendRequestToPolicyServer(t *testing.T) {
-	go startServer()
+	server := httptest.NewServer(defaultHTTPTestingServerHandler)
```
viccuad, line 50: Thanks for this change. I think this should solve the flaky tests!

internal/report/report.go
```diff
@@ -101,6 +103,26 @@ func NewPolicyReport(namespace *v1.Namespace) PolicyReport {
...
+func (r *PolicyReport) CopyResult(result *v1alpha2.PolicyReportResult) {
+	switch result.Result {
+	case StatusFail:
+		r.Summary.Fail++
+	case StatusError:
+		r.Summary.Error++
+	case StatusPass:
+		r.Summary.Pass++
+	}
+	r.Results = append(r.Results, result)
```
viccuad, line 123: Shouldn't we be substituting the old result instead of appending a new one?
Maybe I'm misunderstanding the usage of this function in scanner.go.

We have `CopyResult(result)`, and `AddResult(args that create a result)`.
Maybe `CopyResult(result)` should be named `AddResult(result)`, and `AddResult(args to create result)` should be named `CreateResult(args to create result)`.

internal/report/report.go
```diff
@@ -179,7 +214,106 @@ func (r *ClusterPolicyReport) AddResult( //nolint:dupl
...
+		return false, err
+	}
+	policyResourceVersion, hasPolicyResourceVersion := result.Properties[PropertyPolicyResourceVersion]
+	policyUID, hasPolicyUID := result.Properties[PropertyPolicyUID]
+	return result.Policy == policyName &&
+		hasPolicyResourceVersion && policyResourceVersion == policy.GetResourceVersion() &&
+		hasPolicyUID && types.UID(policyUID) == policy.GetUID(), nil
+}
+
+func findReusableResult(results []*v1alpha2.PolicyReportResult, policy policiesv1.Policy, resource unstructured.Unstructured) *v1alpha2.PolicyReportResult {
```
viccuad, line 258: I would document that for finding the result, we must ensure that:
policy has the same UID and ResourceVersion
resource has the same UID and ResourceVersion
To protect us from incorrectly relaxing this comparison in a future refactor.

internal/report/report.go
```diff
@@ -262,17 +380,15 @@ func newPolicyReportResult( //nolint:funlen,cyclop
 	if policy.IsContextAware() {
 		properties[TypeContextAware] = ValueTypeTrue
 	}
+	// The policy resource version and the policy UID are used to check if the
+	// same result can used in the next scan
```
viccuad, line 384:
```suggestion
	// same result can be reused in the next scan
```

viccuad (changes requested):
LGTM, with minor comments.

The only thing I would like to clarify is the drop of the in-mem store. I would prefer to have both an in-mem store, and a client one, with an interface. This ensures decoupling from client specifics, and the in-mem store can be used for debug and mocks.

In fact, this PR could have been made against the in-mem store, and a later PR could have added the client store with no additional tests needed besides those for the client store.

jvanz:
@viccuad, as we discussed in a call last week, we will keep using the kubernetes go-client cache. Therefore, I'll keep the changes removing our in memory cache and mark your comments around that as resolved.

Besides that, I've made the other changes that you suggested. 

Considering that, bigger changes are approved and you will not be able to review this PR soon, I'll merge it after CI passes. If you see something else that need to be changed, please, let me know. I'll be happy fixing them.

Issue #52, opened 2023-06-14
audit-scanner - MVP Stage 6 - inspect the whole cluster


viccuad:
Change the behavior of the cli app:
* When no flag are specified: scan all the Namespaces and the cluster-wide resources
* When the namespace flag is used: scan only the specified namespace
* When the cluster-wide flag is used: scan only cluster-wide resources
* When the skip-ns list is specified: skip specified list of namespaces

During this phase, we can evaluate each namespace in a sequential way (no parallelization).
    
### Acceptance criteria:
- When no flag is used: scan all the Namespaces and the cluster-wide resources
- When the namespace flag is used: scan only the specified namespace
- When the cluster-wide flag is used: scan only cluster-wide resources
- When the skip-ns list is specified: skip specified list of namespaces
- User cannot use the namespace flag and the cluster-wide one at the same time
- We have unit tests
- We have integration test **if needed**

viccuad:
I consider this should include a config for `audit-scanner`. This config should at leat contain the `--skip-namespaces` value.

viccuad:
I have postponed the config. With the `--ignore-namespaces` implemented with `StringSliceVarP`, we accept multiple flags, and the parsing from CSV is done for us. Hence helm templating is simpler.

PR #53, opened 2023-06-15
feat: Scan all namespaces and cluster by default

viccuad:
## Description

Fix https://github.com/kubewarden/audit-scanner/issues/52

- feat: Adjust flag behaviour
  Not passing any flag: Scan cluster wide and all ns.
  Passing both `--cluster` and `--namespace foo`: error.
- feat: Add new --skipped-ns flag, Add Fetcher.skippedNS
  Add new flag of comma separated list of namespace names to be skipped.
  Add new field to policies.Fetcher{}, Fetcher.skippedNs.
- feat: Add fetcher.GetAuditedNamespaces(), implement scan all ns

## Test

Tested locally. Example:
```
./bin/audit-scanner -l debug -u https://localhost:3000 -p -s "default,kubewarden"
```

### DIFF
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -84,32 +84,25 @@ func Execute() {
 	}
 }
 func startScanner(namespace string, clusterWide bool, scanner Scanner) error {
-	switch clusterWide {
-	case true:
-		if namespace != "" {
-			log.Fatal().Msg("Cannot scan cluster wide and only a namespace at the same time")
-		} else {
-			if err := scanner.ScanClusterWideResources(); err != nil {
-				return err
-			}
-		}
-	case false:
-		if namespace != "" {
-			if err := scanner.ScanNamespace(namespace); err != nil {
-				return err
-			}
-		} else {
-			// neither clusterWide flag nor namespace was provided, default
-			// behaviour of scanning cluster wide and all ns
-			if err := scanner.ScanClusterWideResources(); err != nil {
-				return err
-			}
-			if err := scanner.ScanAllNamespaces(); err != nil {
-				return err
-			}
-		}
+	if clusterWide && namespace != "" {
+		log.Fatal().Msg("Cannot scan cluster wide and only a namespace at the same time")
 	}
-	return nil
+
+	if clusterWide {
+		// only scan clusterwide
+		return scanner.ScanClusterWideResources()
+	}
+	if namespace != "" {
+		// only scan namespace
+		return scanner.ScanNamespace(namespace)
+	}
+
+	// neither clusterWide flag nor namespace was provided, default
+	// behaviour of scanning cluster wide and all ns
+	if err := scanner.ScanClusterWideResources(); err != nil {
+		return err
+	}
+	return scanner.ScanAllNamespaces()
 }
 
 func init() {
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -123,5 +123,5 @@ func init() {
 	rootCmd.Flags().StringP("policy-server-url", "u", "", "Full URL to the PolicyServers, for example https://localhost:3000. Audit scanner will query the needed HTTP path. Useful for out-of-cluster debugging")
 	rootCmd.Flags().VarP(&level, "loglevel", "l", fmt.Sprintf("level of the logs. Supported values are: %v", logconfig.SupportedValues))
 	rootCmd.Flags().BoolVarP(&printJSON, "print", "p", false, "print result of scan in JSON to stdout")
-	rootCmd.Flags().StringP("skipped-ns", "s", "", "Comma separated list of namespace names to be skipped from scan")
+	rootCmd.Flags().StringP("ignore-namespaces", "i", "", "Comma separated list of namespace names to be skipped from scan")
 }
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -81,24 +81,31 @@ func Execute() {
 	}
 }
 func startScanner(namespace string, clusterWide bool, scanner Scanner) error {
-	if clusterWide {
-		if err := scanner.ScanClusterWideResources(); err != nil {
-			return err
+	switch clusterWide {
+	case true:
+		if namespace != "" {
+			log.Fatal().Msg("Cannot scan cluster wide and only a namespace at the same time")
+		} else {
+			if err := scanner.ScanClusterWideResources(); err != nil {
+				return err
+			}
+		}
+	case false:
+		if namespace != "" {
+			if err := scanner.ScanNamespace(namespace); err != nil {
+				return err
+			}
+		} else {
+			// neither clusterWide flag nor namespace was provided, default
+			// behaviour of scanning cluster wide and all ns
+			if err := scanner.ScanClusterWideResources(); err != nil {
+				return err
+			}
+			if err := scanner.ScanAllNamespaces(); err != nil {
+				return err
+			}
 		}
 	}
-
-	if namespace != "" {
-		if err := scanner.ScanNamespace(namespace); err != nil {
-			return err
-		}
-	} else if !clusterWide {
-		// FIXME ScanAllNamespaces is not implemented. Therefore, if we are
-		// scanning cluster wide resource do not trigger this scan. It will failed anyway
-		if err := scanner.ScanAllNamespaces(); err != nil {
-			return err
-		}
-	}
-
 	return nil
 }
 
--- a/cmd/root.go
+++ b/cmd/root.go
@@ -2,6 +2,7 @@ package cmd
 
 import (
 	"fmt"
+	"strings"
 
 	logconfig "github.com/kubewarden/audit-scanner/internal/log"
 	"github.com/kubewarden/audit-scanner/internal/policies"
@@ -55,7 +56,13 @@ There will be a ClusterPolicyReport with results for cluster-wide resources.`,
 			if err != nil {
 				return err
 			}
-			policiesFetcher, err := policies.NewFetcher(kubewardenNamespace)
+			skippedNsCSV, err := cmd.Flags().GetString("skipped-ns")
+			if err != nil {
+				return err
+			}
+			skippedNs := strings.Split(skippedNsCSV, ",")
+
+			policiesFetcher, err := policies.NewFetcher(kubewardenNamespace, skippedNs)
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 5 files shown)


internal/policies/fetcher.go
```diff
@@ -19,20 +21,23 @@ import (
...
 // if both attempts fail.
-func NewFetcher() (*Fetcher, error) {
+func NewFetcher(skippedNs []string) (*Fetcher, error) {
 	client, err := newClient()
 	if err != nil {
 		return nil, err
 	}
 
-	return &Fetcher{client: client, filter: filterAuditablePolicies}, nil
+	return &Fetcher{client: client, skippedNs: skippedNs, filter: filterAuditablePolicies}, nil
```
viccuad, line 40: (opening thread here, but this is about the following)

https://github.com/kubewarden/audit-scanner/blob/248d16d7c3e8dd843b37d6bf87374b91e9b722a5/internal/policies/fetcher.go#L43-L48

We don't need this function. I would like to think about the map-reduce again just in case this informs that we are tackling the problem incorrectly, but I think we can drop it.

cmd/root.go
```diff
@@ -81,24 +88,31 @@ func Execute() {
...
+			}
+		} else {
+			// neither clusterWide flag nor namespace was provided, default
+			// behaviour of scanning cluster wide and all ns
+			if err := scanner.ScanClusterWideResources(); err != nil {
+				return err
+			}
+			if err := scanner.ScanAllNamespaces(); err != nil {
+				return err
+			}
```
jvanz, line 113: My personal preference... I found weird a switch case for a boolean variable.

cmd/root.go
```diff
@@ -109,4 +123,5 @@ func init() {
 	rootCmd.Flags().StringP("policy-server-url", "u", "", "Full URL to the PolicyServers, for example https://localhost:3000. Audit scanner will query the needed HTTP path. Useful for out-of-cluster debugging")
 	rootCmd.Flags().VarP(&level, "loglevel", "l", fmt.Sprintf("level of the logs. Supported values are: %v", logconfig.SupportedValues))
 	rootCmd.Flags().BoolVarP(&printJSON, "print", "p", false, "print result of scan in JSON to stdout")
+	rootCmd.Flags().StringP("skipped-ns", "s", "", "Comma separated list of namespace names to be skipped from scan")
```
jvanz, line 126:
```suggestion
	rootCmd.Flags().StringP("skip-namespaces", "s", "", "Comma separated list of namespace names to be skipped from scan")
```
When using the long CLI flag name, I try to avoid abbreviation.

internal/policies/fetcher.go
jvanz: If we do not neeed it, drop it. :)

jvanz (changes requested):

internal/policies/fetcher.go
```diff
@@ -105,6 +110,24 @@ func (f *Fetcher) GetNamespace(nsName string) (*v1.Namespace, error) {
 	return namespace, nil
 }
 
+// GetAuditedNamespaces gets all namespaces besides the ones in fetcher.skippedNs
+// This function cannot be tested with fake.client, as fake.client doesn't
+// support fields.OneTermNotEqualSelector()
```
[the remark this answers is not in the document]
viccuad: I tried with this test:

```golang
func TestGetAuditedNamespaces(t *testing.T) {
	// it should skip the skippedNs
	c := Fetcher{client: mockClient(), skippedNs: []string{"default"}}
	nsList, err := c.GetAuditedNamespaces()
	if err != nil {
		t.Errorf("error should be nil:  %s", err.Error())
	}

	for _, ns := range nsList.Items {
		if ns.Name == "default" {
			t.Errorf("namespace default was part of skipped ns")
		}
	}
}
```

And I got:

```console
=== RUN   TestGetAuditedNamespaces                                                                                                                                                                                                                                                                                             
{"level":"debug","ns":"default","time":"2023-06-16T08:24:50+02:00","message":"skipping ns"}                                                                                                                                                                                                                                    
    fetcher_test.go:355: error should be nil:  can't list namespaces: field selector ,metadata.name!=default is not in one of the two supported forms "key==val" or "key=val"                                                                                                                                                  
--- FAIL: TestGetAuditedNamespaces (0.00s)  
```

I found an issue talking about it (that I can't find again to link, sigh). But in the code:
https://github.com/kubernetes-sigs/controller-runtime/blob/451530c09937350ff7f5ebf7a00d059b4334e8ee/pkg/client/fake/client.go#L598-L603

cmd/root.go
viccuad: better indeed! applying

cmd/root.go
viccuad: Me too. But the linters complain 3 nested ifs are too much..

viccuad:
rebased on top of main to fix conflicts.

internal/policies/fetcher.go
viccuad: Indeed, done in https://github.com/kubewarden/audit-scanner/pull/53/commits/40df6ef3b4223ab141f5c18a6447c295594ab227

cmd/root.go
viccuad, line 126: I'm not super thrilled of getting the long list of skipped ns as a CSV in a flag. But I don't want to use a config file, and an env var would also consist of a CSV.

cmd/root.go
```diff
@@ -55,7 +56,13 @@ There will be a ClusterPolicyReport with results for cluster-wide resources.`,
...
 				return err
 			}
-			policiesFetcher, err := policies.NewFetcher(kubewardenNamespace)
+			skippedNsCSV, err := cmd.Flags().GetString("skipped-ns")
+			if err != nil {
+				return err
+			}
+			skippedNs := strings.Split(skippedNsCSV, ",")
+
+			policiesFetcher, err := policies.NewFetcher(kubewardenNamespace, skippedNs)
```
flavio, line 65: TODO: look back, can't we just add the kubewarden namespace to the list of the skipped ones?

cmd/root.go
flavio: what about something like that:

```golang
func scanNamespaces(nsNames []string) error {
   nsWithErrors := []string{}
   for namespace in nsNames {
     if err : = scanner.ScanNamespace(namespace); err != nil {
        log.Error(err)
        nsWithErrors = append(nsWithErrors, namespace)
     }
   }
   if somethingWentWrong {
     return fmt.Errorf("Something went wrong while scanning these namespaces, checks the logs for more details: %s", strings.join(nsWithErrors, ", "))
   }
   return nil
}
```

Then we could have the code do something like that:

```
if cli_flags.clusterwide && cli_flags.namespace != "" {
   return error
}

if cli_flags.clusterwide {
   scan_cluster_wide
}

var nsToScan []string
if cli_flags.namespace != "" {
   nsToScan = []string{cli_flags.namespace}
} else {
   nsToScan = find_all_ns()
}

scanNamespaces(nsToScan)
```

cmd/root.go
flavio: nitpicking, maybe we should use "ignore" instead of "skip"

internal/policies/fetcher.go
```diff
@@ -109,6 +105,24 @@ func (f *Fetcher) GetNamespace(nsName string) (*v1.Namespace, error) {
...
+func (f *Fetcher) GetAuditedNamespaces() (*v1.NamespaceList, error) {
+	skipNsFields := fields.Everything()
+	for _, nsName := range f.skippedNs {
+		skipNsFields = fields.AndSelectors(skipNsFields, fields.OneTermNotEqualSelector("metadata.name", nsName))
+		log.Debug().Str("ns", nsName).Msg("skipping ns")
+	}
```
flavio, line 116: Can't you do the equivalent of:

```yaml
  namespaceSelector:
     matchExpressions:
     - key: "kubernetes.io/metadata.name"
       operator: NotIn
       values: [kube-system, cattle-monitoring-system]
```

From my understanding you are building a long `(name != FOO) && (name != BAR) && ...)`

cmd/root.go
viccuad: done in https://github.com/kubewarden/audit-scanner/pull/53/commits/0b6f6f72ceae73ad7226377810b85d887723d731.

cmd/root.go
viccuad: that is indeed more elegant, but there, `find_all_ns()` is right now implemented by `s.policiesFetcher.GetAuditedNamespaces()`. I can consume `policiesFetcher` from the RunE function, but that also adds it to the mock tests.

cmd/root.go
```diff
@@ -55,7 +56,13 @@ There will be a ClusterPolicyReport with results for cluster-wide resources.`,
 			if err != nil {
 				return err
 			}
-			policiesFetcher, err := policies.NewFetcher(kubewardenNamespace)
+			skippedNsCSV, err := cmd.Flags().GetString("skipped-ns")
+			if err != nil {
+				return err
+			}
+			skippedNs := strings.Split(skippedNsCSV, ",")
```
jvanz, line 63: I think you can use [`StringSliceVarP`](https://pkg.go.dev/github.com/spf13/pflag#FlagSet.StringSliceP) instead of `StringP`. It splits the values by comma for you.

Furthermore, it makes more easy to build the yaml in the Helm chart templates. Because we do not need to concat all the skipped namespaces. We can just pass every skipped namespace in a separated flag.

internal/policies/fetcher.go
viccuad: I found no way to do it differently. All functions only accept a string (so, 1 field value), and compare 1 key to 1 field.

One could do `fields.ParseSelector("metadata.name!=" + "default")` and build a SCV of key and fields to select for. I have seen code examples for that in the wild. But looking at how they get parsed in the underlying field code yields the same implementation, plus one then needs to sanitize the namespace names so there can't be an attack when building that string to parse as selector.

cmd/root.go
viccuad: Indeed, better! Done in https://github.com/kubewarden/audit-scanner/pull/53/commits/18d37612e41788eeadeb3bd78ed0c0be85a16b4c.

cmd/root.go
viccuad: refactored even simpler in https://github.com/kubewarden/audit-scanner/pull/53/commits/33ab16402bdcd660536977dddcc338b36222fdc7.

viccuad:
rebased onto main to fix conflicts, ready for review!

cmd/root.go
```diff
@@ -81,25 +84,25 @@ func Execute() {
 	}
 }
 func startScanner(namespace string, clusterWide bool, scanner Scanner) error {
-	if clusterWide {
-		if err := scanner.ScanClusterWideResources(); err != nil {
-			return err
-		}
+	if clusterWide && namespace != "" {
+		log.Fatal().Msg("Cannot scan cluster wide and only a namespace at the same time")
```
jvanz, line 88: Why are these operation mutually exclusive? At first glance, I cannot see any technical reason to block this.

jvanz (approved):

cmd/root.go
viccuad: At some point we may have stated that we only wanted to run one or the other, hence the implementation. I can't find it in the RFCs, so it must have been in an impromptu talk. It comes from the epic task list too.

IMHO there's no problem with it, we want to scan all ns and clusterwide as default anyways.

PR #72, opened 2023-07-21
fix: enable release candidates when required

jvanz:
## Description

Updates the release CI to check if the pushed tag has some text indicating a release candidate and set properly the flag to publish the release as final release or a candidate.

Related to https://github.com/kubewarden/policy-server/pull/497/files#r1268541197


viccuad (approved):
LGTM!

PR #88, opened 2023-08-07
deps: Pin wg-policy-prototypes

viccuad:
## Description

Upstream is untagged, and we need to consume the PolicyReports v1alpha2.
Follow-up of https://github.com/kubewarden/audit-scanner/pull/86

Hopefully this will make the bot not try to bump that dep.

## Test

CI


flavio (approved):
Good choice :+1:

PR #106, opened 2023-09-05
chore(deps): be explicit about the GH actions versions

flavio:
Provide full version for GH actions


fabriziosestito (approved):
LGTM, double-checked

(and google sent me to a captcha page thinking I was a bot)

PR #109, opened 2023-09-08
fix: Always update PolicyReport, even if scanned resources number is 0

viccuad:
## Description

Fix https://github.com/kubewarden/kubewarden-controller/issues/516

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,7 +5,7 @@ on:
   push:
   pull_request:
 
-# Declare default permissions as read only.
+    # Declare default permissions as read only.
 permissions: read-all
 
 jobs:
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -5,7 +5,7 @@ on:
   push:
   pull_request:
 
-# Declare default permissions as read only.
+    # Declare default permissions as read only.
```
jvanz, line 8:
```suggestion
# Declare default permissions as read only.
```

jvanz (approved):

.github/workflows/ci.yml
viccuad: woops! thanks!

flavio (approved):

PR #150, opened 2023-12-05
Adds in-mem `PolicyReportStore` implementation (#107)

### LINKED ISSUE #107: Feature request: Add in-mem `PolicyReportStore{}` implementation
Requested by @nnelas, see https://github.com/kubewarden/audit-scanner/issues/102.

Currently, audit-scanner uses this [PolicyReportStore{}](https://github.com/kubewarden/audit-scanner/blob/main/internal/report/store.go#L27-L30) , which uses a k8s client and therefore uses Kubernetes (Etcd) as cache for the PolicyReports.

This store cache is used to see [historical data of audits here](https://github.com/kubewarden/audit-scanner/blob/main/internal/scanner/scanner.go#L141-L154). This means that if nothing has changed in the audit request condition, the audit-scanner then [skips reevaluating that specific case here](https://github.com/kubewarden/audit-scanner/blob/main/internal/scanner/scanner.go#L298-L308).

### Acceptance criteria

1. Create a `store` Go interface that satisfies the current store implementation at [report/store.go ](https://github.com/kubewarden/audit-scanner/blob/main/internal/report/store.go).
1. Refactor current `report/store.go` into `report/store-kubernetes.go`, as an implementation of the store interface.
   - Check that its unit tests pass.
   - Consume the new `store` Go interface in `scanner/scanner.go`, defaulting to the `store-kubernetes` one.
1. Add a new `memory` store with a slice of [PolicyReports{}](https://github.com/kubewarden/audit-scanner/blob/main/internal/report/report.go#L21-L23), each with their [PolicyReportResults{}](https://github.com/kubewarden/audit-scanner/blob/main/internal/report/report.go#L24C1-L26C2). This new memory store could be under another file `report/store-memory.go` for example.
    -  Unit tests for the new store are provided if needed. Given that this is just operating on in-mem arrays, maybe there's not much to write here.
    - This store will then provide the cache functionality and allow for skipping of reevaluation as expected.
1. Provide an optional cli flag, `--store TYPE`, with possible types `memory`, `kubernetes`. Defaults to `--store kubernetes` for ease on first Kubewarden installations.
2. Expose this functionality by adding a new field in the `kubewarden-controller` helm chart.

nnelas:
## Description

Fix #107

As specified in the original issue, this commit:
- Create a store Go interface that satisfies the current store implementation at `report/store.go`
- Refactor current `report/store.go` into `report/store_kubernetes.go`, as an implementation of the store interface
- Adds a new memory store with 2 maps, one with `ClusterPolicyReports` and another one with `PolicyReports`
- Provide an optional cli flag, `--store TYPE`, with possible types memory, kubernetes. Defaults to --store kubernetes for ease on first Kubewarden installations.

Besides that, this commit also:
- Changes `printJSON` to `outputScan` to make it more aligned with what the input params is
- `outputScan` only outputs at the end of the flow, to prevent having a lot of logs whenever someone scans the entire cluster (we were outputting the entire store on each namespace loop)

It also seems that running `make fmt` introduced a couple of unwanted changes that I'm happy to revert to keep this commit smaller. Finally, I also had to do a couple of tweaks to make golangci lint happy.

## Test

To test this pull request, you can run the following commands:

```shell
make unit-tests
```

### Potential improvement

One thing that we can also do in the future is to parse (in some way) the `toJSON()` result and instead of logging the entire report in a single log block, we could have a log line per violation. What do you think?

Also, I'll be out-of-office for the rest of the week, but I'll return next Monday. Looking forward to seeing your comments! ❤️


internal/report/store.go
```diff
@@ -1,291 +1,45 @@
...
-	"k8s.io/apimachinery/pkg/labels"
-	"k8s.io/apimachinery/pkg/types"
-	"k8s.io/client-go/kubernetes/scheme"
-	"k8s.io/client-go/util/retry"
-	ctrl "sigs.k8s.io/controller-runtime"
-	"sigs.k8s.io/controller-runtime/pkg/client"
-	polReport "sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1alpha2"
+const (
+	KUBERNETES string = "KUBERNETES"
+	MEMORY     string = "MEMORY"
```
jvanz, line 5:
```suggestion
	KUBERNETES string = "kubernetes"
	MEMORY     string = "memory"
```

jvanz (changes requested):
I could not finish testing the PR yet. But I'll continue soon.

internal/report/store_memory.go
```diff
@@ -0,0 +1,198 @@
...
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
+		// get the latest report version to be updated
+		latestReport, err := s.GetPolicyReport(report.GetNamespace())
+		if err != nil {
+			return err
+		}
+		latestReport.Summary = report.Summary
+		latestReport.Results = report.Results
+		return s.UpdatePolicyReport(&latestReport)
+	})
```
jvanz, line 138: I think you can remove the `RetryOnConflict`. This is use in K8s report because of the conflict it can happen when updating the report in the backend. Here, I don't expect that to happen so often. Therefore, checking the `GetPolicyReport` error once is enough.

internal/report/store_memory.go
```diff
@@ -0,0 +1,198 @@
...
+	if getErr != nil {
+		// Create new Policy Report if not found
+		if errors.Is(getErr, constants.ErrResourceNotFound) {
+			// Update will create a new one if it doesn't exist
+			return s.UpdateClusterPolicyReport(report)
+		}
+		return getErr
+	}
+	// Update existing Policy Report
+	retryErr := retry.RetryOnConflict(retry.DefaultRetry, func() error {
```
jvanz, line 157: I think you can remove the `RetryOnConflict`. This is use in K8s report because of the conflict it can happen when updating the report in the backend. Here, I don't expect that to happen so often. Therefore, checking the `GetPolicyReport` error once is enough.

cmd/root.go
```diff
@@ -27,63 +29,92 @@ type Scanner interface {
...
+		scanner, err := scanner.NewScanner(policyReportStore, policiesFetcher, resourcesFetcher, insecureSSL, caCertFile)
+		if err != nil {
+			return err
+		}
+
+		if err := startScanner(namespace, clusterWide, scanner); err != nil {
+			return err
+		}
+
+		if outputScan {
```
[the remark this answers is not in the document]
viccuad: I think it should stay. The in memory store should be there just for testing. We implement the interface in the memory store or for example a future SQL backend.

README.md
```diff
@@ -10,10 +10,13 @@
...
 
-The results of the scan are made available via `PolicyReport` objects. Each Namespace
+The results of the scan can be made available via `PolicyReport` objects. Each Namespace
 has its own dedicated `PolicyReport`. Cluster-wide resources compliance is available via
 the `ClusterPolicyReport` resource. 
 
+Instead of relying on `PolicyReport` objects, one can also configure Audit scanner to
+save all this information in-memory only, by specifying `--store MEMORY`.
+
 # Deployment
```
viccuad, line 20: (opening thread here randomly)

> One thing that we can also do in the future is to parse (in some way) the toJSON() result and instead of logging the entire report in a single log block, we could have a log line per violation. What do you think?

At that point, I think it would be better  to expose the reports with proper structured logging.

cmd/root.go
```diff
@@ -27,63 +29,92 @@ type Scanner interface {
...
+		caCertFile, err := cmd.Flags().GetString("extra-ca")
+		if err != nil {
+			return err
+		}
+		storeType, err := cmd.Flags().GetString("store")
+		if err != nil {
+			return err
+		}
+
+		policyReportStore, err := getPolicyReportStore(storeType)
```
viccuad, line 75: We should be creating the store in scanner.go as we were doing before, and not in `cmd/root.go`. Ideally, `cmd/root.go` should be only used for reading the cmd flags, and calling an audit-scanner lnternal library with all the logic. Right now `cmd/root.go` already does more than it should (create policiesFetcher, resourcesFetcher, etc) but that should change in the future.

Refactoring into an internal lib would simplify testing and mocks, yet it is out of the scope for this PR. But that doesn't mean we shouldn't keep piling up things in `cmd/root.go`.

internal/report/store.go
```diff
@@ -1,291 +1,45 @@
...
-		client.MatchingLabels(map[string]string{
-			LabelAppManagedBy: LabelApp,
-		}))
-	if err != nil {
-		return err
-	}
-	return nil
-}
+	// UpdateClusterPolicyReport updates a ClusterPolicyReport or PolicyReport.
+	// ns argument is used in case of namespaced PolicyReport
```
viccuad, line 26: there is no ns argument, nor this function updates a PolicyReport

internal/report/store_memory.go
```diff
@@ -0,0 +1,198 @@
...
+)
+
+type MemoryPolicyReportStore struct {
+	// prCache is a map between a namespace and a PolicyReport
+	prCache map[string]PolicyReport
+
+	// cprCache is a map between a name and a ClusterPolicyReport
+	cprCache map[string]ClusterPolicyReport
+
+	mutex sync.RWMutex
```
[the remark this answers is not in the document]
viccuad: I think we should have them, we can't ensure it's write and read is concurrent safe.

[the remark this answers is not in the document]
viccuad: True (yet they don't hurt now that they are there, so for me it's not a merge blocker to remove them).

internal/report/store.go
```diff
@@ -1,291 +1,45 @@
...
+	UpdateClusterPolicyReport(report *ClusterPolicyReport) error
```
viccuad, line 27: Since only `SaveClusterPolicyReport` and `SavePolicyReport` are used from outside of the package, we should not have `UpdateClusterPolicyReport` and `UpdatePolicyReport` as part of the interface, and they should be private on the implemented packages.

internal/report/store_memory_test.go
```diff
@@ -0,0 +1,173 @@
+package report_test
+
+import (
+	"github.com/stretchr/testify/require"
```
viccuad, line 4: I would prefer not to add a (dev) dependency if it isn't needed, maybe we can do something else besides `require.NoError()`.

internal/report/store_memory.go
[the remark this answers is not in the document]
nnelas: I think I was just suffering from anticipation, I will remove those locks since we don't need them yet

internal/report/store.go
nnelas: changed! :)

internal/report/store_memory.go
nnelas: that makes sense, thank you for noticing that! changed for `SavePolicyReport` and [SaveClusterPolicyReport](https://github.com/kubewarden/audit-scanner/pull/150#discussion_r1417270463)

internal/report/store.go
nnelas: to be honest, I had mixed feelings about this one, I did notice that these 2 methods were not being used outside of the package but I didn't want to change the interface since you could've left them public for future work. in this case, I'll remove them from the interface

internal/report/store.go
nnelas: although I should've been more careful when moving the original godocs, this method will be removed with https://github.com/kubewarden/audit-scanner/pull/150#discussion_r1419164395

internal/report/store.go
nnelas: also, with that in mind, does it make sense to remove both `RemovePolicyReport` and `RemoveAllNamespacedPolicyReports` given that we currently have no usage for them?

README.md
nnelas: yeap, completely agree with that. but perhaps it can be another feature request with another PR, or do you think it's feasible to include that implementation under this one?

internal/report/store_memory.go
```diff
@@ -0,0 +1,198 @@
...
+}
+
+func (s *MemoryPolicyReportStore) UpdatePolicyReport(report *PolicyReport) error {
+	s.mutex.Lock()
+	defer s.mutex.Unlock()
+
+	s.prCache[report.GetNamespace()] = *report
+
+	summary, _ := report.GetSummaryJSON()
+	log.Info().
```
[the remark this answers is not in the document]
nnelas: agreed! I've pushed a new commit that just align these log levels and their messages between `PolicyReports`, `ClusterPolicyReports` and policy stores

internal/report/store_memory_test.go
nnelas: make sense, I ended up using a plain nil-check for errors and removed `require` as a dependency

internal/policies/fetcher.go
```diff
@@ -194,7 +194,7 @@ func (f *Fetcher) getAdmissionPolicies(namespace string) ([]policiesv1.Admission
 	return policies.Items, nil
 }
 
-func newClient() (client.Client, error) { //nolint:ireturn
+func newClient() (client.Client, error) { //nolint:ireturn // no specific client type
```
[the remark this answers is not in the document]
nnelas: to be fair, this was a quick and dirty "solution" to try to keep `ireturn` lint happy. for some reason, it's a bit flaky as sometimes `nolintlint` complains that that comment is not needed but it's not consistent :/

cmd/root.go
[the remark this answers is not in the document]
nnelas: my apologies @viccuad, but I didn't understood what you meant with
> I think we should have this logic to do ToJSON and print in cmd/root.go, and it should be on scanner.go.

internal/report/store.go
```diff
@@ -1,291 +1,45 @@
...
-		Dict("dict", zerolog.Dict().
-			Str("report name", report.GetName()).
-			Str("report ns", report.GetNamespace()).
-			Str("report resourceVersion", report.GetResourceVersion()).
-			Str("summary", summary),
-		).Msg("updated PolicyReport")
-	return nil
-}
+// PolicyReportStore caches the latest version of PolicyReports
+// There is 2 different stores available:
```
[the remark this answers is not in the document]
nnelas: my goal here was to have an easy way for any client of PolicyReportStore API to know which implementations are available and what's their goal, but you're right, doesn't make sense to have an API knowing about it's implementations

cmd/root.go
nnelas: I completely agree with you, in this case, I went to the easy path and decided to add another dependency to `cmd/root.go` when shouldn't. I've refactored this bit with your suggestions. there's still `outputScan` there that I'm happy to move it elsewhere after https://github.com/kubewarden/audit-scanner/pull/150#discussion_r1417210469

nnelas:
hallo @jvanz and @viccuad ! first of all, thank you so much for your suggestions/comments, really helpful to keep the codebase consistent and aligned with your standards ❤️ 

I think I addressed all your topics, except for https://github.com/kubewarden/audit-scanner/pull/150#discussion_r1417210469 and https://github.com/kubewarden/audit-scanner/pull/150#discussion_r1417221641 since I need further help/clarification for them

internal/report/store_memory.go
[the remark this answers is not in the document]
viccuad: I'm ok with them being removed for now :).

cmd/root.go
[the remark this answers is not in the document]
viccuad: I meant that I would like the block `if outputScan { scanner.ReportStore.ToJSON() }` to stay in `scanner/scanner.go` as it was before. This puts all the logic in the internal libs, and leaves `cmd/*` thinner, just dealing with flags and making the big calls. It will simplify future testing and refactoring too.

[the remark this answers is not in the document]
nnelas: I get your point now, thank you for your explanation @viccuad ! one thing that it's important to mention is that this way we will output all store data (either Kubernetes or Memory) multiple times per namespace which is also not ideal

internal/policies/fetcher_test.go
```diff
@@ -357,20 +361,22 @@ var nsDefault = v1.Namespace{
...
+
 var nsKubewarden = v1.Namespace{
 	ObjectMeta: metav1.ObjectMeta{
 		Name:   "kubewarden",
 		Labels: map[string]string{"kubernetes.io/metadata.name": "kubewarden"},
 	},
 }
 
-func mockClient(initObjs ...k8sClient.Object) k8sClient.Client { //nolint:ireturn
+func mockClient(initObjs ...k8sClient.Object) k8sClient.Client { //nolint:ireturn // no specific client type
```
jvanz, line 379: Again, try the trick about the linter that I've shared in another comment

internal/report/store.go
jvanz: Yeah, if we are not using, I'm fine removing these methods. I think that  code no in use should not be in the code base.

jvanz (changes requested):
Thanks @nnelas! I think we are almost merging this PR :heart: 

I just have some small requests about the linter. Besides that, I would say we can wait for another @viccuad review and merge if he approved.

jvanz:
Furthermore, @nnelas can you rebase and fix the conflicts in the `go.sum` file? :grin:

internal/report/store_memory.go
```diff
@@ -0,0 +1,172 @@
...
+// that uses in-memory cache to store `PolicyReports` and `ClusterPolicyReports`
+type MemoryPolicyReportStore struct {
+	// prCache is a map between a namespace and a PolicyReport
+	prCache map[string]PolicyReport
+
+	// cprCache is a map between a name and a ClusterPolicyReport
+	cprCache map[string]ClusterPolicyReport
+}
+
+func NewMemoryPolicyReportStore() (*MemoryPolicyReportStore, error) {
```
flavio, line 26: Since it never fails, we could omit `error` from the return type

README.md
flavio: I too think it would be better to get this done with another PR

flavio (commented):
Great contribution, thanks a lot @nnelas!

internal/report/store.go
nnelas: completely agree @jvanz , I'll remove those `Remove*` methods

jvanz (approved):
LGTM. @kubewarden/kubewarden-developers remember to squash the commits when merging this PR!

internal/report/store_memory.go
nnelas: I had mixed feelings about this one, I was thinking of keeping all constructors for all `PolicyReportStore` implementations coherent, but I'm now thinking that doesn't make sense because we'll have different dependencies for each store. I'll remove the `error`

internal/policies/fetcher.go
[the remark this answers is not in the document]
nnelas: thank you so much for your tip! It did the trick 🎉

go.mod
```diff
@@ -8,14 +8,15 @@ require (
 	github.com/kubewarden/kubewarden-controller v1.9.0
 	github.com/rs/zerolog v1.31.0
 	github.com/spf13/cobra v1.8.0
+	golang.org/x/exp v0.0.0-20231206192017-f3f8817b8deb
 	k8s.io/api v0.29.0
 	k8s.io/apimachinery v0.29.0
 	k8s.io/client-go v0.29.0
 	sigs.k8s.io/controller-runtime v0.16.3
-	sigs.k8s.io/wg-policy-prototypes v0.0.0-20230505033312-51c21979086a
```
viccuad, line 15: This represents a bump in the wg-policy-prototypes that changes the field of the crds. See the comparison here:
https://github.com/kubernetes-sigs/wg-policy-prototypes/compare/51c21979086a654f60d5d65a95f9681e954816ac..22764d64b0c3f79d2293b67000b4d9ebca197623

Particularly the file `policy-report/crd/v1beta1/wgpolicyk8s.io_clusterpolicyreports.yaml`. There's only 2 new fields to the CRDs, and no deletions, but that means that PolicyReports created with this new library version may not be backwards compatible.

Thing is, https://github.com/kubernetes-sigs/wg-policy-prototypes is not correctly tagged nor has proper releases, and other community consumers keep just developing the future v1beta1 (listed as [WIP](https://github.com/kubernetes-sigs/wg-policy-prototypes/tree/master/policy-report#policy-report)), so we need to be careful.

I would like to postpone this bump.

nnelas: oh, I see! that makes sense, let me revert that change in a bit

internal/report/store_kubernetes.go
viccuad, line 1: (for future PRs if there's need to move files around (like store.go -> store_kubernetes.go) I would prefer the usage of `git mv` to create a distinct commit for it (easier to review, and maintains original authorship).

viccuad (commented):
Looks good! Many thanks for dealing with the changes!

The only blocker to me is the bump on the `sigs.k8s.io/wg-policy-prototypes` library, I would like to postpone that.

internal/report/store_memory_test.go
viccuad: Many thanks!

internal/report/store_kubernetes.go
nnelas: didn't know about that! thanks for the TIL 🙇

nnelas:
> Looks good! Many thanks for dealing with the changes!
> 
> The only blocker to me is the bump on the `sigs.k8s.io/wg-policy-prototypes` library, I would like to postpone that.

hallo @viccuad ! thank you so much for your review 🙇 I've reverted that upgrade under https://github.com/kubewarden/audit-scanner/pull/150/commits/10c4eee528da06b71fc00241f1cb3836b705e0d5 so I think I've addressed all your suggestions. in the meantime, I'll create a PR in https://github.com/kubewarden/helm-charts to use this new `--store` flag

viccuad (approved):
LGTM, thanks!

viccuad:
golangci-lint was again cryptically failing, and passing locally. Ended up deleting its cache under https://github.com/kubewarden/audit-scanner/actions/caches.

GH says merging is blocked because all commits must be signed, yet [DCO reports all are signed](https://github.com/kubewarden/audit-scanner/pull/150/checks?check_run_id=19681233535) Oo. Maybe it got confused with unsigned merge commits from gh.

@flavio could you merge this please?

flavio (approved):

flavio:
Squashed and merged. Thanks for your contribution @nnelas!

Issue #119, opened 2023-09-15
Feature Request: use `scrach` as container image foundation


flavio:
### Is your feature request related to a problem?

Currently the audit-scanner image is built on top of distroless. We don't need all the stuff provided by distroless, we can switch to `scratch`

### Solution you'd like

Have the final container image based on `scratch`

PR #152, opened 2023-12-06
feat: Build image with scratch instead of distroless

viccuad:
## Description

Fix https://github.com/kubewarden/audit-scanner/issues/119

## Test

Tested locally by running `ghcr.io/viccuad/test/audit-scanner:scratch ` in a cluster.

Checked `syft ghcr.io/viccuad/test/audit-scanner:scratch -vv`, which uses the go-module-binary-cataloger to find the deps:
```
[0001] DEBUG discovered 63 packages cataloger=go-module-binary-cataloger
(...)
```

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -17,11 +17,14 @@ COPY internal/ internal/
 # Build
 RUN CGO_ENABLED=0 GOOS=linux GO111MODULE=on go build -a -o audit-scanner .
 
-# Use distroless as minimal base image to package the audit-scanner binary
-# Refer to https://github.com/GoogleContainerTools/distroless for more details
-FROM gcr.io/distroless/static:nonroot
-WORKDIR /
-COPY --from=builder /workspace/audit-scanner .
-USER 65532:65532
+FROM alpine AS cfg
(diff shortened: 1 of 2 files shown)


Dockerfile
```diff
@@ -17,11 +17,14 @@ COPY internal/ internal/
...
+FROM alpine AS cfg
+RUN echo "audit-scanner:x:65533:65533::/tmp:/sbin/nologin" >> /etc/passwd
+RUN echo "audit-scanner:x:65533:audit-scanner" >> /etc/group
 
+# Copy the statically-linked binary into a scratch container.
+FROM scratch
+COPY --from=cfg /etc/passwd /etc/passwd
+COPY --from=cfg /etc/group /etc/group
+COPY --from=builder --chmod=0755 /workspace/audit-scanner /audit-scanner
+USER 65532:65532
```
viccuad, line 29: tried by also doing `COPY go.mod go.sum /`, to see if syft would use that instead of the signatures in the binary go executable, but it didn't, so I left it out for now.

jvanz (approved):

PR #162, opened 2023-12-19
ci: Fix artifact upload and download

viccuad:
## Description

{upload,download}-artifact v4 don't accept uploading the same named artifact. Suffix their names with the arch, and download all of them at once into the same dir, as it was before.

Relates to https://github.com/kubewarden/audit-scanner/pull/160.

## Test

Tested on my fork: https://github.com/viccuad/audit-scanner/actions/runs/7261395579/job/19782800832


flavio (approved):
Thanks for the fix

PR #165, opened 2023-12-20
feat: adds logging per `(Cluster)PolicyReport` result

### LINKED ISSUE #102: Feature Request: Allow configuration of scanning reports output
### Is your feature request related to a problem?

Hallo! 👋 

I'm currently analysing the hypothesis of using the Kubewarden stack as the policy manager for a multi-tenant cluster, and I'm struggling a bit with `audit-scanner` as it requires by default the instances of a (Cluster)PolicyReport for `PolicyReportStore`. From our experience with Gatekeeper, working with clusters with hundreds of namespaces, all these CRD instances led to increased pressure on `apiserver` and `etcd` nodes, which if possible would be something we would want to avoid.

### Solution you'd like

As a result, I would like to understand if it would be possible to add a configuration that would define where would `audit-scanner` output/write a scanning report. To keep compatibility, it could keep writing to a (Cluster)PolicyReport by default. It could also output as JSON to the stdout (just like the `--print` flag that you currently have) or even write to a file on a volumeMount.

### Alternatives you've considered

Looking at `audit-scanner` source-code, I couldn't find any alternative without changing its source code.

### Anything else?

Please let me know if this makes sense and if it's aligned with the vision that you have for `audit-scanner`. I can also happily contribute with a PR, if this is something that you can't fit in your roadmap in the foreseeable future.

Thank you so much for all the hard work that you've put into this project! 🙌

nnelas:
## Description

Fix #102

Instead of having a big log line containing all the content from the PolicyStore (which includes the `ClusterPolicyReport` and all the namespaced `PolicyReports`), this commit creates a `PolicyReportLogger` which logs all the relevant information per scanned resource.

## Test

I didn't create any tests for this class because it's just responsible for logging. However, I did test this manually by running:
```
./bin/audit-scanner --store memory --policy-server-url "https://localhost:8443" --insecure-ssl --output-scan
```

eg.:
```
{"level":"info","dict":{"policy":"cap-service-external-ip","rule":"service-external-ip","result":"pass","message":"","resource_api_version":"v1","resource_kind":"Service","resource_namespace":"some-namespace","resource_name":"some-pod","resource_version":"780797213"},"time":"2023-12-20T15:20:58Z"}

{"level":"info","dict":{"policy":"cap-pod-disruption-budget","rule":"pod-disruption-budget","result":"fail","message":"minAvailable isn't allowed in a PDB","resource_api_version":"policy/v1","resource_kind":"PodDisruptionBudget","resource_namespace":"some-namespace","resource_name":"some-pod","resource_version":"438425862"},"time":"2023-12-20T15:20:58Z"}
```

### Potential improvement

I've added the fields from `(Cluster)PolicyReports` that we need but happy to iterate this PR to include more information that you also find relevant!

### DIFF
--- a/internal/log/policy_report_logger.go
+++ b/internal/log/policy_report_logger.go
@@ -1,0 +1,67 @@
+package log
+
+import (
+	"github.com/kubewarden/audit-scanner/internal/report"
+	"github.com/rs/zerolog"
+	"github.com/rs/zerolog/log"
+)
+
+type PolicyReportLogger struct {
+}
+
+// LogClusterPolicyReport will create a log line per each cluster-wide scanned resource
+func (l *PolicyReportLogger) LogClusterPolicyReport(report *report.ClusterPolicyReport) {
+	log.Info().
+		Dict("dict", zerolog.Dict().
+			Int("pass", report.Summary.Pass).
+			Int("fail", report.Summary.Fail).
+			Int("warn", report.Summary.Warn).
+			Int("error", report.Summary.Error).
+			Int("skip", report.Summary.Skip)).
+		Msg("ClusterPolicyReport summary")
+
+	for _, result := range report.Results {
+		log.Info().
+			Dict("dict", zerolog.Dict().
+				Str("policy", result.Policy).
+				Str("rule", result.Rule).
+				Str("result", string(result.Result)).
+				Str("message", result.Description).
+				// although subjects is a list, there's only 1 element
+				Str("resource_api_version", result.Subjects[0].APIVersion).
+				Str("resource_kind", result.Subjects[0].Kind).
+				Str("resource_namespace", result.Subjects[0].Namespace).
+				Str("resource_name", result.Subjects[0].Name).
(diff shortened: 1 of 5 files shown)


flavio (approved):
LGTM, waiting for reviews from other people who are more familiar with this area of the codebase

viccuad (approved):
LGTM, thanks!

viccuad:
@flavio I see the merging blocked, can you have a look at the DCO app? The commit has the correct `Signed-off-by` and DCO app passes, but the merging is blocked. This is the second time it happens, but on my PRs it's not happening.

nnelas:
> @nnelas this is blocked because we require all commits to be signed with gpg. I'm going to override the check this time, can you configure your git client to sign its commits (see [here](https://docs.github.com/en/authentication/managing-commit-signature-verification/signing-commits))

another TIL! I thought that `git commit -s` would suffice. I've already changed my git client for future contributions. thank you @flavio and @viccuad ❤️

Issue #202, opened 2024-02-23
Switch to resource-based Reports


flavio:
Currently audit-scanner creates one PolicyReport per Namespace and one ClusterPolicyReport per cluster. These reports contain the compliance status of ALL the resources inside of a Namespace or the compliance status of all the cluster-wide resources.

On clusters with many resources this leads to the generation of really big PolicyReports. These resources are so big that they cannot even be saved inside of etcd. Meaning that, after the scanner is run, some PolicyReports are missing.

We need to generate many smaller PolicyReport resources, one per resource defined inside of the cluster.

With the new format, the UX will improve. Currently, finding the non-compliant resources is complicated, because you have to search for this information inside of a big report. With the new approach, it's easier to find the non-compliant resources:

```console
$ kubectl get polr -n kube-system -o wide
NAME                                   KIND         NAME                                         PASS   FAIL   WARN   ERROR   SKIP   AGE
049a4ec1-32a5-4417-9184-1a59cfaa1ca6   DaemonSet    kindnet                                      9      3      0      0       0      16m
049d2cca-c30f-4f26-a70a-dfcc2cc5f433   DaemonSet    kube-proxy                                   9      3      0      0       0      16m
1d491ec4-ca84-4b3a-960a-a2aefa3219ba   Pod          kube-controller-manager-kind-control-plane   10     2      0      0       0      16m
34fa05b8-40cc-4bd3-836e-077abf4c126e   Pod          kindnet-qtq54                                9      3      0      0       0      16m
3997d5d0-363a-4820-8768-4be3788b3968   Pod          kube-proxy-tcgcz                             9      3      0      0       0      16m
4434c0ac-e27f-41eb-b4c2-b1a7aca8056a   ReplicaSet   coredns-5dd5756b68                           12     0      0      0       0      16m
487df031-11d8-4ab4-b089-dfc0db1e533e   Pod          kube-apiserver-kind-control-plane            10     2      0      0       0      16m
553c0601-b995-4ed8-a36b-11e7cb38893b   Pod          kube-proxy-jdsck                             9      3      0      0       0      16m
89044d72-8a1e-4af0-877b-9be727dc3ec4   Pod          kindnet-7rrns                                9      3      0      0       0      16m
9eb8c5c0-fe5c-4c7d-96c3-3ff65c361f4f   Pod          etcd-kind-control-plane                      10     2      0      0       0      16m
[...]

PR #204, opened 2024-02-26
refactor: remove label selector grouping of resources

fabriziosestito:
## Description

This PR removes the label selector grouping of resources in favor of matching the label selectors locally.
This is a prerequisite for fixing #202 and removing the chance to fetch duplicated resources if a lot of policy label selectors overlap.
It also fixes the skipped/auditable policy count.

## Tradeoff
There is an edge case where we could fetch more resources than needed. For instance when a user has defined a policy matching just one resource of many, we will fetch all the resources and apply the label selector to each one, since we are not using server-side filtering anymore.
However, by using the pagination, we do not keep all the resources in memory, and in general, it is better to optimize the code for the most common scenario.


internal/policies/client.go
```diff
@@ -193,10 +169,13 @@ func (f *Client) getAdmissionPolicies(ctx context.Context, namespace string) ([]
...
 
-// groupPoliciesByGVRAndLabelSelectorg groups policies by GVR and LabelSelector.
+// groupPoliciesByGVRAndLabelSelectorg groups policies by GVR.
 // If namespaced is true, it will skip cluster-wide resources, otherwise it will skip namespaced resources.
-func (f *Client) groupPoliciesByGVRAndLabelSelector(ctx context.Context, policies []policiesv1.Policy, namespaced bool) (map[schema.GroupVersionResource]map[string][]*Policy, error) {
-	resources := make(map[schema.GroupVersionResource]map[string][]*Policy)
+func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.Policy, namespaced bool) (*Policies, error) { //nolint:funlen
+	policiesByGVR := make(map[schema.GroupVersionResource][]*Policy)
+	auditablePolicies := map[string]struct{}{}
+	skippedPolicies := map[string]struct{}{}
```
flavio, line 177: Can we replace the usage of maps with a [set](https://github.com/deckarep/golang-set)? Just to make things more explicit

internal/policies/client.go
```diff
@@ -328,3 +313,24 @@ func (f *Client) getServiceByAppLabel(ctx context.Context, appLabel string, name
...
 	return &serviceList.Items[0], nil
 }
+
+func isAuditable(policy policiesv1.Policy) bool {
+	return policy.GetBackgroundAudit() &&
+		policy.GetStatus().PolicyStatus == policiesv1.PolicyStatusActive &&
+		isCreateActionPresentWithoutAllResources(policy)
+}
+
+func isCreateActionPresentWithoutAllResources(policy policiesv1.Policy) bool {
```
flavio, line 323: nit: add some codedoc explaining the purpose of the function

flavio (approved):
LGTM, I like this change, it makes the code easier to grasp.

I've left some minor comments

internal/policies/client.go
fabriziosestito: we could but we would need to introduce a new dependency

jvanz (approved):

PR #216, opened 2024-03-04
feat: add resource-based reports cache 

### LINKED ISSUE #194: [SURE-7630] audit-scanner gets killed by OOM
### Current Behavior

Given the following environment:
- 150 Deployments, resulting in 300 Pods
- 14 enabled policies

The audit-scanner Pod gets killed by the OOM killer because it consumes more than 2Gb of memory

### Expected Behavior

audit-scanner completes without being killed by OOM

### Steps To Reproduce

1. Use latest version of audit-scanner
2. Create 150 deployments of nginx, with 2 replicas each
3. Run audit-scanner

### Environment

```markdown
- Kubewarden >= 1.8.0
```

### Anything else?

Related issues:

- [x] https://github.com/kubewarden/audit-scanner/pull/201
- [x] https://github.com/kubewarden/audit-scanner/issues/202
- [x] https://github.com/kubewarden/docs/issues/365
- [x] https://github.com/rancher/kubewarden-ui/issues/609
- [x] Update release notes: explain the users they have to delete the old PolicyReports and the old ClusterPolicyReport manually

fabriziosestito:
## Description

Fix #194 

This PR adds back the  re-use of policy report results, if the resource and the policy were not changed from the previous audit, so that we avoid hitting the policy server to get the same result.

## Test

Unit tests have been added.

### DIFF
--- a/internal/report/constants.go
+++ b/internal/report/constants.go
@@ -3,7 +3,7 @@ package report
 const (
 	policyReportSource            = "kubewarden"
 	propertyPolicyResourceVersion = "policy-resource-version"
-	PropertyPolicyUID             = "policy-uid"
+	propertyPolicyUID             = "policy-uid"
 )
 
 const (
--- a/internal/report/report.go
+++ b/internal/report/report.go
@@ -1,8 +1,6 @@
 package report
 
 import (
-	"time"
-
 	policiesv1 "github.com/kubewarden/kubewarden-controller/pkg/apis/policies/v1"
 	admissionv1 "k8s.io/api/admission/v1"
 	corev1 "k8s.io/api/core/v1"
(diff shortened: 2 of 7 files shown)


fabriziosestito:
@kubewarden/kubewarden-developers 
We found a problem with the original implementation and [RFC](https://github.com/kubewarden/rfc/blob/main/rfc/0011-audit-checks.md#step-5-attempt-to-reuse-previous-evaluation-results). The design does not take into account that results of policies that make use of host capabilities cannot be cached, since they are not idempotent. 

Even if it is possible to detect if a policy is context-aware, at the moment it is not possible to know if a policy is using host capabilities in general, therefore we cannot be sure if we can reuse the result or not while auditing a resource.
Moreover, by switching to resource-based reports, reusing results means hitting the Kubernetes API server every time we audit a resource (vs retrieving the report once like it was done in the old implementation).

flavio:
Closing, we're not going to implement this feature as explained by @fabriziosestito

PR #220, opened 2024-03-11
feat: parallel audit requests

fabriziosestito:
## Description

This PR implements parallel audit requests to the policy server.
The parallelism is currently set to 100 requests, but in the future, we could expose this value to the user through the CLI so that it can be tuned. 

### Gnu time output
Before the fix on a cluster with ~300 resources and 15 policies:

```
Percent of CPU this job got: 79%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:41.24
```

After the fix
```
Percent of CPU this job got: 109%
Elapsed (wall clock) time (h:mm:ss or m:ss): 0:30.42
```

In the case of slower policies (for instance policies using context-aware capabilities) is safe to assume that the elapsed time difference should be higher.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -55,6 +55,7 @@ require (
 	golang.org/x/exp v0.0.0-20240112132812-db7319d0e0e3 // indirect
 	golang.org/x/net v0.20.0 // indirect
 	golang.org/x/oauth2 v0.16.0 // indirect
+	golang.org/x/sync v0.6.0 // indirect
 	golang.org/x/sys v0.16.0 // indirect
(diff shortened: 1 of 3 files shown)


flavio (approved):
Nice!

flavio:
This is going to be part of the final 1.11 release

CC @brunorene

flavio:
This fixes https://github.com/kubewarden/audit-scanner/pull/220

viccuad (approved):
LGTM!

viccuad:
This feels more a feature than a fix, for it appear as such in the GH release it is enough to change the PR title.

PR #225, opened 2024-03-14
chore: configure code coverage

fabriziosestito:
## Description

Tune code coverage settings


fabriziosestito:
no change in coverage after ignoring packages, closing

Issue #233, opened 2024-03-25
Audit scanner errors out instead of skipping `*` policies


kravciak:
The example is safelabels policy - if I leave `*` defaults in rules audit scanner jobs fail with error.
As a side-effect cluster has many `Error` pods from failed cronjobs which are not cleaned automatically.

```yaml
apiVersion: policies.kubewarden.io/v1
kind: ClusterAdmissionPolicy
metadata:
  name: starlabels
  namespace: ''
  annotations:
    artifacthub/pkg: safe-labels/safe-labels/0.1.14
spec:
  backgroundAudit: true
  policyServer: default
  module: ghcr.io/kubewarden/policies/safe-labels:v0.1.14
  rules:
    - apiGroups:
        - '*'
      apiVersions:
        - '*'
      resources:
        - '*'
      operations:
        - CREATE
        - UPDATE
  mutating: false
  namespaceSelector:
    matchExpressions:
      - key: kubernetes.io/metadata.name
        operator: NotIn
        values:
          - calico-system
          - cattle-alerting
          - cattle-fleet-local-system
          - cattle-fleet-system
          - cattle-global-data
          - cattle-global-nt
          - cattle-impersonation-system
          - cattle-istio
          - cattle-logging
          - cattle-pipeline
          - cattle-prometheus
          - cattle-system
          - cert-manager
          - ingress-nginx
          - kube-node-lease
          - kube-public
          - kube-system
          - rancher-operator-system
          - security-scan
          - tigera-operator
    matchLabels: {}
  settings:
    description: null
    constrained_labels: {}
    denied_labels:
      - qqq
    mandatory_labels:
      - aaa
  contextAwareResources: []
  mode: monitor

```

```bash
~ k logs -n cattle-kubewarden-system audit-scanner-28522836-ggpkr
{"level":"info","time":"2024-03-25T12:36:00Z","message":"clusterwide resources scan started"}
{"level":"fatal","error":"no matches for */*, Resource=*","time":"2024-03-25T12:36:00Z","message":"Error on cmd.Execute()"}
```

TechDufus:
I'm seeing this exact behavior as well:

```json
{"level":"info","time":"2024-03-25T12:00:20Z","message":"clusterwide resources scan started"}
{"level":"fatal","error":"no matches for */*, Resource=*","time":"2024-03-25T12:00:20Z","message":"Error on cmd.Execute()"}
```

PR #235, opened 2024-03-27
fix: wildcard resources not being skipped

fabriziosestito:
## Description

Fix: #233

### DIFF
--- a/internal/policies/client.go
+++ b/internal/policies/client.go
@@ -171,12 +171,19 @@ func (f *Client) getAdmissionPolicies(ctx context.Context, namespace string) ([]
 
 // groupPoliciesByGVRAndLabelSelectorg groups policies by GVR.
 // If namespaced is true, it will skip cluster-wide resources, otherwise it will skip namespaced resources.
-func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.Policy, namespaced bool) (*Policies, error) { //nolint:funlen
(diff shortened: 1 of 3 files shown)


internal/policies/client.go
```diff
@@ -204,17 +211,20 @@ func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.P
...
+
+							continue
+						}
 
-							log.Debug().Str("policy", policy.GetUniqueName()).
-								Bool("backgroundAudit", policy.GetBackgroundAudit()).
-								Bool("active", policy.GetStatus().PolicyStatus == policiesv1.PolicyStatusActive).
-								Bool("create", isCreateActionPresentWithoutAllResources(policy)).
-								Msg("not auditable policy, skipping!")
+						if policy.GetStatus().PolicyStatus != policiesv1.PolicyStatusActive {
```
viccuad, line 221: this could also be moved outside of the `rules` for , akin to `!isCreateActionPresentWithoutAllResources(policy)`.

fabriziosestito: the reason why it is here is because we need the check the above condition if the policy targets resource with the right scope, otherwise policies that are not active but targeting for instance namespaced resource when scanning a namespace will be considered as skipped as well

viccuad: then, maybe we need a `break`? If not, it seems to be that we will append the policy several times per resource and version.

jvanz:
:tada:

PR #274, opened 2024-05-15
fix: delete old reports after policy is deleted.

### LINKED ISSUE #263: Reports are not updated after last policy removal
I have a CA policy that produces audit-report on the namespace.
When I delete this policy audit report on namespace is not removed / retriggered, leaving behind clusterpolicyreport.

## To reproduce:

- install Kubewarden with recommended policies
- create safe-labels policy with following settings
```yaml
apiVersion: policies.kubewarden.io.v1
kind: ClusterAdmissionPolicy
metadata:
  name: safelabels
spec:
  backgroundAudit: true
  policyServer: default
  module: ghcr.io/kubewarden/policies/safe-labels:v0.1.14
  rules:
    - apiGroups:
        - ''
      apiVersions:
        - ''
      resources:
        - 'namespaces'
      operations:
        - CREATE
        - UPDATE
  settings:
    constrained_labels: {}
    denied_labels:
      - qqq
```
- add denied label to default namespace `kubectl label ns default qqq=1`
- trigger audit scan (at this point clusterpolicyreport for default namespace with 1 FAIL is created as expected)
- delete `safelabels` policy and trigger audit scan

#### Current state:
Clusterpolicyreport for default namespace with 1 FAIL is not removed, UI points to non-existent policy

#### Expexted state:
Clusterpolicyreport results from non-existent policies are cleaned

jvanz:
## Description

When the report has only result from policies that are not deployed in the cluster anymore, the audit scanner leaves the reports there until the resource is deleted. This is not right. This commit changes this behaviour deleting all the reports that contains only audit result from policies removed from the cluster.
    
Furthermore, this commit updates the scheme package changing how the known types are registered by using functions available from the policies and wgpolicies packages that does this.

Fix #263 

## Test

```shell
make unit-tests
```


internal/scanner/scanner.go
```diff
@@ -102,25 +102,69 @@ func NewScanner(
 	}, nil
 }
 
+func getNamespacePolicyReportsNames(ctx context.Context, k8sClient *k8s.Client, nsName string) (map[string]interface{}, error) {
```
viccuad, line 105: why not return a set of strings instead of what we are doing right now?
I fail to see the benefit of returning a `map[string]interface{}`, where we set the value of the key to `nil`.

internal/scheme/scheme.go
```diff
@@ -1,38 +1,22 @@
...
 	"k8s.io/apimachinery/pkg/runtime"
-	"k8s.io/apimachinery/pkg/runtime/schema"
 	"k8s.io/client-go/kubernetes/scheme"
 	wgpolicy "sigs.k8s.io/wg-policy-prototypes/policy-report/pkg/api/wgpolicyk8s.io/v1alpha2"
 )
 
-func NewScheme() *runtime.Scheme {
+func NewScheme() (*runtime.Scheme, error) {
 	scheme := scheme.Scheme
+	err := policiesv1.AddToScheme(scheme)
```
viccuad, line 12: I think this allows us to drop several unused constants from internal/constants/constants.go now.

viccuad (commented):
Looks good, some minor quibbles.

fabriziosestito (changes requested):
Thanks for this! There is some clean-up that can be done in the clients, but generally, this solution does the job.  
However, we could try a different approach that does not require pulling all the reports from the API server, as this is an expensive operation.

We could add two methods to the store, something like:

```go
func (s *PolicyReportStore) DeleteOldPolicyReports(ctx context.Context, runID string) error {
	labelSelector, err := labels.Parse(fmt.Sprintf("audit-scanner-run-id!=%s", runID))
	if err != nil {
		return err
	}

	return s.client.DeleteAllOf(ctx, &wgpolicy.PolicyReport{}, &client.DeleteAllOfOptions{ListOptions: client.ListOptions{
		LabelSelector: labelSelector,
	}})
}
```

and then generate an `uuid` on each run. 

The uuid can be used in the `report.NewPolicyReport` and `NewClusterPolicyReport` constructors to set the `audit-scanner-run-id` label.

If a report contains only results of policies that were deleted, the report will be not touched and the uuid will not be updated.

Therefore, we could safely delete those reports that do not match the labelselector, with the method above, at the end of the scanner loop.

I think this results in much less code and we avoid hitting the API server, which could be slow, especially if the user has a lot of resources being audited.

LMK what you think.

jvanz:
> LMK what you think.

I liked the proposed solution. I'll do that.

fabriziosestito (approved):
LGTM
Just one nit: we use both `runUID`/`run-uid` and `scanUID`/`scan-uid` naming in tests data and variables. Could we change them to be consistent (`runUID`/`run-uid` everywhere)?

internal/scanner/scanner_test.go
```diff
@@ -233,30 +252,39 @@ func TestScanAllNamespaces(t *testing.T) {
...
-	err = scanner.ScanAllNamespaces(context.Background())
+
+	runUID := uuid.New().String()
+	err = scanner.ScanAllNamespaces(context.Background(), runUID)
 	require.NoError(t, err)
 
 	policyReport := wgpolicy.PolicyReport{}
 
+	err = client.Get(context.TODO(), types.NamespacedName{Name: oldPolicyReport.GetName(), Namespace: oldPolicyReport.GetNamespace()}, oldPolicyReport)
+	require.ErrorContains(t, err, "not found")
```
fabriziosestito, line 263: Maybe import `apimachineryErrors "k8s.io/apimachinery/pkg/api/errors"`

then
`require.True(t, apimachineryErrors.IsNotFound(err))`

internal/scanner/scanner_test.go
```diff
@@ -375,7 +404,9 @@ func TestScanClusterWideResources(t *testing.T) {
 
 	scanner, err := NewScanner(policiesClient, k8sClient, policyReportStore, false, false, true, "")
 	require.NoError(t, err)
-	err = scanner.ScanClusterWideResources(context.Background())
+
```
fabriziosestito, line 407: Could we add an old report here too, so we are sure that we are testing the deletion of `ClusterPolicyReports` too?

jvanz: Ops, thanks! I forgot to update this test.

PR #282, opened 2024-05-27
fix: make audit-scanner send requests to all PolicyServer replicas

flavio:
By dafault, the Go http client reuses connections. This casuses scaling issues when a PolicyServer instance is backed by multiple replicas. In this scanerio, the requests are sent to the samePolicyServer Pod, causing the load to be unevenly distributed.To avoid this, we disable disable keep-alives, which ensures anew connection is created for each evaluation request.

I was able to reproduce this issue on a local cluster with a PollicyServer instance backed by 3 replicas.

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -46,6 +46,8 @@ type Scanner struct {
 // If insecureClient is false, it will read the caCertFile and add it to the in-app
 // cert trust store. This gets used by the httpClient when connection to
 // PolicyServers endpoints.
+//
+//nolint:funlen // the comment lines make this check fail
 func NewScanner(
 	policiesClient *policies.Client,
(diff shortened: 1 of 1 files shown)


internal/scanner/scanner.go
```diff
@@ -82,6 +82,15 @@ func NewScanner(
 	if !ok {
 		return nil, errors.New("failed to build httpClient: failed http.Transport type assertion")
 	}
+
+	// By dafault, the http client reuses connections. This casuses
```
fabriziosestito, line 86: typo, default

fabriziosestito (approved):
LGTM just a typo in the comment

internal/scanner/scanner.go
flavio: fixed, thanks

PR #286, opened 2024-05-29
parallelize everything

flavio:
Change how the audit process is done, allowing more parallilazion if wanted by the user.

This PR adds these new flags (I'm open to change their names, I'm not super fond of them):

- `--parallel-namespaces`: how many Namespaces can be audited in parallel. Default is 1.
- `--parallel-resources`: how many resources can be audited in parallel. Default is 50.
- `--parallel-policies`: when auditing a specic resource, how many policies can be invoked in parallel. Default is 10.

By tuning these values, the user has direct control over how many HTTP requests are sent to the policy server(s) at the same time.
For example, `--parallel-namespaces 2 --parallel-resources 100 --parallel-policies 10` translates to a maximum of `2 * 100 * 10 = 2000` HTTP requests at the same time.

## Benchmarks

Given the following scenario:

- 30k RoleBinding objects, all defined inside of the `default` namespace
- 1 Policy Server, with 3 replicas
- 4 policies targeting RoleBinding objects

### No parallelization

```bash
audit-scanner --disable-store --parallel-namespaces 1 --parallel-policies 1 --parallel-resources 1
```

Leads to the following results:

```
User time (seconds): 77.85
System time (seconds): 43.11
Percent of CPU this job got: 29%
Elapsed (wall clock) time (h:mm:ss or m:ss): 6m 50.45s
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 209872
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 0
Minor (reclaiming a frame) page faults: 10814
Voluntary context switches: 2569133
Involuntary context switches: 983543
Swaps: 0
File system inputs: 0
File system outputs: 0
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

### Parallelization

```bash
./audit-scanner --disable-store --parallel-namespaces 2 --parallel-policies 10 --parallel-resources 100
```

Leads to the following results:

```
User time (seconds): 150.32
System time (seconds): 33.26
Percent of CPU this job got: 60%
Elapsed (wall clock) time (h:mm:ss or m:ss): 5m 3.90s
Average shared text size (kbytes): 0
Average unshared data size (kbytes): 0
Average stack size (kbytes): 0
Average total size (kbytes): 0
Maximum resident set size (kbytes): 392304
Average resident set size (kbytes): 0
Major (requiring I/O) page faults: 0
Minor (reclaiming a frame) page faults: 157699
Voluntary context switches: 464957
Involuntary context switches: 785465
Swaps: 0
File system inputs: 0
File system outputs: 0
Socket messages sent: 0
Socket messages received: 0
Signals delivered: 0
Page size (bytes): 4096
Exit status: 0
```

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -35,10 +35,11 @@ type Scanner struct {
 	k8sClient         *k8s.Client
 	policyReportStore *report.PolicyReportStore
 	// http client used to make requests against the Policy Server
-	httpClient              http.Client
-	outputScan              bool
-	disableStore            bool
-	parallelResourcesAudits int
+	httpClient               http.Client
+	outputScan               bool
+	disableStore             bool
+	parallelNamespacesAudits int
+	parallelResourcesAudits  int
 }
 
 // NewScanner creates a new scanner
@@ -55,6 +56,7 @@ func NewScanner(
 	disableStore bool,
 	insecureClient bool,
 	caCertFile string,
+	parallelNamespacesAudits int,
 	parallelResourcesAudits int,
 ) (*Scanner, error) {
 	// Get the SystemCertPool to build an in-app cert pool from it
@@ -104,13 +106,14 @@ func NewScanner(
 	}
 
 	return &Scanner{
-		policiesClient:          policiesClient,
-		k8sClient:               k8sClient,
-		policyReportStore:       policyReportStore,
-		httpClient:              httpClient,
-		outputScan:              outputScan,
-		disableStore:            disableStore,
-		parallelResourcesAudits: parallelResourcesAudits,
+		policiesClient:           policiesClient,
+		k8sClient:                k8sClient,
+		policyReportStore:        policyReportStore,
+		httpClient:               httpClient,
+		outputScan:               outputScan,
+		disableStore:             disableStore,
+		parallelNamespacesAudits: parallelNamespacesAudits,
+		parallelResourcesAudits:  parallelResourcesAudits,
 	}, nil
 }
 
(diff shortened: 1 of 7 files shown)


internal/scanner/scanner.go
```diff
@@ -290,46 +298,99 @@ func (s *Scanner) ScanClusterWideResources(ctx context.Context, runUID string) e
...
+	semaphore := semaphore.NewWeighted(int64(s.parallelPoliciesAudits))
+	var workers sync.WaitGroup
 
-	for _, p := range policies {
-		url := p.PolicyServer
-		policy := p.Policy
+	auditResults := make(chan policyAuditResult)
+	reportsDone := make(chan bool)
+
+	go func() {
```
fabriziosestito, line 324: do we need this go routine here?
couldn't we just 
```go
for res := range auditResults {
		report.AddResultToPolicyReport(res.policyReport, res.policy, res.admissionReview, res.errored)
	}
```

after the workers have completed the auditing?

jvanz (commented):
I have mixing feelings about this 3 cli flags. Knowing the code, I understand and they make sense. But thinking in the user UX I think this can be improved. Maybe we can have a simpler CLI flags something like `max-parallel-request` and `worker-pool-size`. Where the first controls the number of request going out of the scanner. While the second, controls the number of internal go routines doing the scanner logic. Or maybe have only the `max-parallel-request` and making everything in the scanner running in parallel, because the scanner logic is not that complex and the most expensive operation are the requests. 

But I'm not sure if we want to do that now (if we decided to do it in the first place) or in a future iteration (v1.14.0)... Besides that, I'm fine with the code if we want to keep the current approach from this PR.

internal/scanner/scanner.go
flavio: Done

fabriziosestito (approved):
:shipit:

jvanz (approved):

PR #298, opened 2024-06-17
fix: structured logs should not have spaces inside of keys

flavio:
The keys of structured logs should not have spaces inside of them. That makes querying the logs harder.


flavio:
CC @brunorene

viccuad (approved):

Issue #307, opened 2024-06-25
Audit Scanner fails completely with typo in policy


Martin-Weiss:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

Using this policy (example from https://artifacthub.io/packages/kubewarden/verify-image-signatures/verify-image-signatures)

```
apiVersion: policies.kubewarden.io/v1
kind: ClusterAdmissionPolicy
metadata:
  name: verify-image-signatures
spec:
  module: ghcr.io/kubewarden/policies/verify-image-signatures:v0.1.7
  rules:
  - apiGroups: ["", "apps", "batch"]
    apiVersions: ["v1"]
    resources: ["pods", "deployments", "statefulsets", "replicationcontrollers", "jobs", "cronjobs"]
    operations:
    - CREATE
    - UPDATE
  mutating: true
  settings:
    signatures:
    - image: "*"
      githubActions:
        owner: "kubewarden"
        repo: "app-example" 
```

 with kubewarden-controller-2.1.0 kubewarden-crds-1.5.1 kubewarden-defaults-2.0.3  ends up in audit-scanner failing fatal with the following

```
kubectl -n kubewarden logs audit-scanner-28655400-49zwv 
{"level":"info","RunUID":"b87b30fc-5633-42c8-ab21-bf585d6826a6","time":"2024-06-25T14:00:01Z","message":"clusterwide resources scan started"}
{"level":"fatal","error":"no matches for apps/v1, Resource=pods","time":"2024-06-25T14:00:02Z","message":"Error on cmd.Execute()"}
```

### Expected Behavior

Audit scanner must not fail completely if a policy is not perfect ok.

### Steps To Reproduce

See above - more details here:https://github.com/Martin-Weiss/rancher-fleet/tree/bb68ca7288b42dcc7ddea27d5ffabde7fe241d22/kubewarden

### Environment

```markdown
rke2 1.28.10 x86_64 on linux
```

viccuad:
Hi, there's several problems here.

## The reproduced is misconfigured:

- The policy module is an old tag [v0.1.7, from 2022-08](https://github.com/kubewarden/verify-image-signatures/releases/tag/v0.1.7),  and this policy version is incompatible with the provided `spec.settings`.
- The `spec.rules` is incorrect and ambiguous (how does k8s know that apiGroups
  `batch` applies only to `cronjobs` and not `pods` for example?).

At first view, one could think that the reproducer was crafted from the
[...]

PR #312, opened 2024-07-03
fix: Skip policies with unknown GVR or policy-server URL

viccuad:
## Description

Fix https://github.com/kubewarden/audit-scanner/issues/307

For policies that target an unknown GVR, or for which no valid
policy-server URL can be constructed, emit error logs and count them as
errored. Continue with the audit scan.

This is a regression fix from previous audit-scanner 1.11 refactor.

For that, add new `Policies.ErroredNum` field. This errored tally will get
printed in the logs, but can't be printed in (Cluster)PolicyReports, as there's
no report created for an unknown unexistent resource.

Note that policies that target an unknown GVR may be incorrectly configured,
or may have been correctly configured against a CRD, but the CRD has been removed by now.

## Test

Added unit tests, CI.

## Additional Information

With the current (Cluster)PolicyReport data structures, it isn't possible to save an error on them for policies that target an unknown GVR.

### DIFF
--- a/internal/policies/client.go
+++ b/internal/policies/client.go
@@ -34,8 +34,10 @@ type Policies struct {
 	PoliciesByGVR map[schema.GroupVersionResource][]*Policy
 	// PolicyNum represents the number of policies
 	PolicyNum int
-	// SkippedNum represents the number of skipped policies
+	// SkippedNum represents the number of skipped policies that don't match audit constraints
 	SkippedNum int
+	// ErroredNum represents the number of errored policies. These policies may be misconfigured
+	ErroredNum int
 }
 
 // Policy represents a policy and the URL of the policy server where it is running
@@ -171,10 +173,12 @@ func (f *Client) getAdmissionPolicies(ctx context.Context, namespace string) ([]
 
 // groupPoliciesByGVRAndLabelSelectorg groups policies by GVR.
 // If namespaced is true, it will skip cluster-wide resources, otherwise it will skip namespaced resources.
+// If the policy targets an unknown GVR or the policy server URL cannot be constructed, the policy will be counted as errored.
 func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.Policy, namespaced bool) (*Policies, error) { //nolint:funlen
 	policiesByGVR := make(map[schema.GroupVersionResource][]*Policy)
 	auditablePolicies := map[string]struct{}{}
 	skippedPolicies := map[string]struct{}{}
+	erroredPolicies := map[string]struct{}{}
 
 	for _, policy := range policies {
 		rules := filterWildcardRules(policy.GetRules())
@@ -201,14 +205,17 @@ func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.P
 
 		groupVersionResources, err := f.getGroupVersionResources(rules, namespaced)
 		if err != nil {
-			return nil, err
+			erroredPolicies[policy.GetUniqueName()] = struct{}{}
+			log.Error().Err(err).Str("policy", policy.GetUniqueName()).Msg("failed to obtain unknown GroupVersion resources. The policy may be misconfigured, skipping as error...")
+			continue
 		}
+
 		if len(groupVersionResources) == 0 {
 			log.
 				Debug().
 				Str("policy", policy.GetUniqueName()).
 				Bool("namespaced", namespaced).
-				Msg("the policy does not target resources with the selected scope")
+				Msg("the policy does not target resources within the selected scope")
 
 			continue
 		}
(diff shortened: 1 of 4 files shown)


internal/policies/client.go
```diff
@@ -229,11 +236,12 @@ func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.P
 
 		url, err := f.getPolicyServerURLRunningPolicy(ctx, policy)
 		if err != nil {
-			return nil, err
+			erroredPolicies[policy.GetUniqueName()] = struct{}{}
+			log.Error().Err(err).Str("policy", policy.GetUniqueName()).Msg("failed to obtain matching policy-server URL, skipping as error...")
+			continue
```
flavio, line 241: Shouldn't these belong to "skipped"? They are properly configured (meaning they reference valid GVR), but they are not yet active

internal/policies/client.go
```diff
@@ -171,10 +173,12 @@ func (f *Client) getAdmissionPolicies(ctx context.Context, namespace string) ([]
 
 // groupPoliciesByGVRAndLabelSelectorg groups policies by GVR.
 // If namespaced is true, it will skip cluster-wide resources, otherwise it will skip namespaced resources.
+// If the policy targets an unknown GVR or the policy server URL cannot be constructed, the policy will be counted as errored.
 func (f *Client) groupPoliciesByGVR(ctx context.Context, policies []policiesv1.Policy, namespaced bool) (*Policies, error) { //nolint:funlen
 	policiesByGVR := make(map[schema.GroupVersionResource][]*Policy)
 	auditablePolicies := map[string]struct{}{}
 	skippedPolicies := map[string]struct{}{}
+	erroredPolicies := map[string]struct{}{}
```
flavio, line 181: `errored` can be a bit misleading, since we do keep track of policies that "error" (we send a POST request, but we get back something that isn't a 200 response).

In this case I would call them "misconfigured"

internal/policies/client.go
fabriziosestito: +1, I think these should go to skipped

internal/scanner/scanner.go
```diff
@@ -142,19 +142,22 @@ func (s *Scanner) ScanNamespace(ctx context.Context, nsName, runUID string) erro
...
+		log.Error().Err(err).Str("namespace", nsName).Msg("failed to obtain auditable policies")
 		return err
 	}
+
 	log.Info().
 		Str("namespace", nsName).
 		Dict("dict", zerolog.Dict().
 			Int("policies-to-evaluate", policies.PolicyNum).
-			Int("policies-skipped", policies.SkippedNum),
+			Int("policies-skipped", policies.SkippedNum).
```
fabriziosestito, line 153: we need to do the same for https://github.com/kubewarden/audit-scanner/blob/c841bc1d6c438db00cbf6317708a04bc8b227dca/internal/scanner/scanner.go#L255

internal/policies/client.go
viccuad: No, in this case, we already know the policy is active from the previous [if block](https://github.com/kubewarden/audit-scanner/blob/main/internal/policies/client.go#L223-L228), and we are failing to construct a valid policy-server URL.

internal/scanner/scanner.go
viccuad: True! done in https://github.com/kubewarden/audit-scanner/pull/312/commits/8d4afbbe0e5249ee2707701858692033fa629e29.

internal/policies/client.go
viccuad: I could create `erroredPolicies`, `misconfiguredPolicies`, but there's no certain way to know if policies are misconfigured, hence I prefer to be verbose. I see 3 possibilities:

1. A true error for policy-server URL, see https://github.com/kubewarden/audit-scanner/pull/312#discussion_r1665309373.
2. A misconfigured policy for the GVR issue, because `spec.rules` is incorrect. Technically misconfigured, I would like to count that as error for visibility. There's no easy way to ascertain this from 3.
3. A policy that was created for a CRD that isn't present anymore. Technically misconfigured. I find the chance of this happening is lesser than 2.

internal/policies/client.go
flavio: You're right, leave them as errored. This is really a corner case

internal/policies/client.go
flavio: All right, let's keep them into errored, otherwise we will end up with lots of possible outcomes -> confuse the end user.

@fabriziosestito do you agree with that?

fabriziosestito: @flavio @viccuad I am ok with this

viccuad: resolving then!

fabriziosestito (approved):

Issue #334, opened 2024-07-30
Errored policies with exceeded deadline are being recorded as `fail` instead


viccuad:
### Is there an existing issue for this?

- [X] I have searched the existing issues

### Current Behavior

With a policy that errors. For example because it exceeds the deadline, such as `ghcr.io/kubewarden/policies/user-group-psp:v0.6.0` with `settings.validate_container_image_configuration: true`:

```yaml
apiVersion: policies.kubewarden.io/v1
kind: ClusterAdmissionPolicy
metadata:
  annotations:
    io.kubewarden.policy.category: PSP
    io.kubewarden.policy.severity: medium
    meta.helm.sh/release-name: kubewarden-defaults
    meta.helm.sh/release-namespace: kubewarden
  name: do-not-run-as-root
spec:
  backgroundAudit: true
  mode: monitor
  module: ghcr.io/kubewarden/policies/user-group-psp:v0.6.0
  mutating: true
  policyServer: default
  rules:
  - apiGroups:
    - ""
    apiVersions:
    - v1
    operations:
    - CREATE
    resources:
    - pods
  settings:
    run_as_group:
      rule: RunAsAny
    run_as_user:
      rule: MustRunAsNonRoot
    supplemental_groups:
      rule: RunAsAny
    validate_container_image_configuration: true
```

The run of this policy for `audit` in the policy-server ends in error:

```
2024-07-30T11:42:05.815220Z ERROR request{method=POST uri=https://policy-server-default.kubewarden.svc:8443/audit/clusterwide-do-not-run-as-root version=HTTP/2.0}:audit{host="policy-server-default-5545c65fb9-zkqhk" policy_id="clusterwide-do-not-run-as-root" kind="Pod" kind_group="" kind_version="v1" name="nginx-with-se
crets" namespace="default" operation="CREATE" request_uid="24c1e904-110c-46da-b106-2625159a5aca" resource="Pod" resource_group="" resource_version="v1" subresource=""}:validate{self=PolicyEvaluator { runtime: "wapc" } settings={"run_as_group": Object {"rule": String("RunAsAny")}, "run_as_user": Object {"rule": String("
MustRunAsNonRoot")}, "supplemental_groups": Object {"rule": String("RunAsAny")}, "validate_container_image_configuration": Bool(true)}}: wasmtime_provider: Failure invoking guest module handler: error while executing at wasm backtrace:
    0: 0x174348 - user_group_psp.wasm!__rust_alloc
[...]

PR #336, opened 2024-07-31
fix: Record errored AdmissionReview.Response as errors and not failures

viccuad:
## Description

Fix https://github.com/kubewarden/audit-scanner/issues/334

Previously, we recorded
`AdmissionReviewRequest` that errored on send as `error`,
and `AdmissionReviewResponse` with return code != 200 as `failure`.

Now, if an `AdmissionReviewResponse` has a return code != 200 (typically, 500),
record it as `error` instead of `failure`.

If an AdmissionReviewResponse is errored (Response.Result.Code != 200),
it must have a human-readable message in `Response.Result.Message`;
use it on the PolicyReportResult.

Also, AdmissionReviewResponse that reject will have that message, keep
using it.

## Test

Expanded 1 test. CI.


viccuad:
Tested succesfully against a real cluster with a policy that has a `context deadline exceeded` error, besides normal failing and passing policies.

jvanz (approved):

fabriziosestito (approved):

Issue #337, opened 2024-07-31
Add tally of skipped policies to the (Cluster)PolicyReport `spec.Summary.Skip`


viccuad:
Right now filter the policies that apply to a specific resource when we do the audit. We generate a tally of skipped policies, which is printed in logs.
E.g: https://github.com/kubewarden/audit-scanner/blob/ad93d59393308f0e798671199e0d1d22a7dad5f9/internal/scanner/scanner.go#L151

### Acceptance criteria

- Pass that tally into the resulting (Cluster)PolicyReport
- Recheck that the policies.erroredNum is actually being populated (this one [here](https://github.com/kubewarden/audit-scanner/blob/ad93d59393308f0e798671199e0d1d22a7dad5f9/internal/scanner/scanner.go#L152)). With https://github.com/kubewarden/audit-scanner/pull/336, we started counting correctly the errored policies, which can only be known when one exercises the policy. Hence this tally may be superfluous in that log.

PR #339, opened 2024-08-01
fix: Propagate skipped,errored numbers from audited policies

viccuad:
## Description

Fix  https://github.com/kubewarden/audit-scanner/issues/337

When we get all auditable policies for a namespace (or clusterwide) we
have a number of skipped policies (those skipped because of constraints
like no * in GVR) and a number of errored policies (unknown GVR for
example).

Propagate those numbers to the creation of the (Cluster)PolicyReport
for its `spec.Summary`.

Note: when we audit resources later on, we can have errored or skipped
policies, those get added to `spec.Summary` at that time.

## Test

CI

### DIFF
--- a/internal/scanner/scanner.go
+++ b/internal/scanner/scanner.go
@@ -152,7 +152,7 @@ func (s *Scanner) ScanNamespace(ctx context.Context, nsName, runUID string) erro
 			Int("policies-errored", policies.ErroredNum),
 		).Msg("policy count")
 
-	for gvr, policies := range policies.PoliciesByGVR {
+	for gvr, pols := range policies.PoliciesByGVR {
 		pager, err := s.k8sClient.GetResources(gvr, nsName)
 		if err != nil {
 			log.Error().Err(err).Str("gvr", gvr.String()).Str("ns", nsName).Msg("failed to get resources")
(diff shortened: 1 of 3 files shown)


internal/scanner/scanner.go
```diff
@@ -257,7 +257,7 @@ func (s *Scanner) ScanClusterWideResources(ctx context.Context, runUID string) e
 			Int("parallel-resources-audits", s.parallelResourcesAudits),
 		).Msg("cluster admission policies count")
 
-	for gvr, policies := range policies.PoliciesByGVR {
+	for gvr, policiesSlice := range policies.PoliciesByGVR {
```
fabriziosestito, line 260: Personally, I would stick with the old name, as idiomatic go does not use Hungarian notation and I find it good enough.

internal/report/report.go
```diff
@@ -73,7 +73,7 @@ func AddResultToPolicyReport(
 }
 
 // NewClusterPolicyReport creates a new ClusterPolicyReport from a given resource
-func NewClusterPolicyReport(runUID string, resource unstructured.Unstructured) *wgpolicy.ClusterPolicyReport {
+func NewClusterPolicyReport(runUID string, resource unstructured.Unstructured, skipppedPoliciesNum, erroredPoliciesNum int) *wgpolicy.ClusterPolicyReport {
```
fabriziosestito, line 76: we could leave the constructor as is and use the fields directly

internal/scanner/scanner.go
viccuad: This is to not confuse the `policies *Policies{}` with the `policies []Policy` , that is already selected and gets used in the goroutine later on. If not, both identifiers clash.

I could rename `policiesSlice` to `pols` to just use inside the for.

It could be possible to pass `policies *Policies{}` , `resourceGVR` to `auditResource()`, and that way we would have also `policies.Skipped` and so on. But then we are passing the whole hashmap of policies on each `auditResource()` call, incrementing the mem usage.

viccuad: Renamed to `pols`.

internal/report/report.go
viccuad: Indeed, dunno what I was thinking. Changed, which removes quite the unneeded changes.

fabriziosestito (approved):

PR #371, opened 2024-09-17
fix: correct logic for retrieving cluster admission policies targeting a specific namespace

fabriziosestito:
This PR fixes the logic for getting ClusterAdmissionPolicies for a specific namespace.

After [fixing the namespace selector in the controller](https://github.com/kubewarden/kubewarden-controller/pull/883) and updating the dependency, some tests started failing.

The old selector was modifying the namespace selector by adding a NotIn condition, which caused the code to always return namespaces for ClusterAdmissionPolicies when the selector was empty. Once the getter was fixed, an empty selector led to an empty namespace list, breaking the logic.

This PR changes the logic by first fetching all ClusterAdmissionPolicies and then filtering out the ones that don't match the target namespace's label selector, if any.


flavio (approved):
Thanks for the fix and for having improved the code!

Issue #366, opened 2024-09-10
Support PolicyGroups in Audit Scanner


viccuad:
Part of https://github.com/kubewarden/audit-scanner/issues/366.

### Acceptance criteria

- Add (Cluster)PolicyGroups to the list of policies that are considered in audit scans. This doesn't include saving the messages in `Warnings` of `AdmissionReviews` as part of the PolicyReportResults.
- Add unit tests.

PR #373, opened 2024-09-17
feat: support PolicyGroups

fabriziosestito:
## Description

Fixes: #366, supersedes: #370

### DIFF
--- a/internal/policies/client.go
+++ b/internal/policies/client.go
@@ -62,22 +62,38 @@ func NewClient(client client.Client, kubewardenNamespace string, policyServerURL
 }
 
 // GetPoliciesByNamespace gets all the auditable policies for a given namespace.
-func (f *Client) GetPoliciesByNamespace(ctx context.Context, namespace string) (*Policies, error) {
+func (f *Client) GetPoliciesByNamespace(ctx context.Context, namespace *corev1.Namespace) (*Policies, error) {
 	var policies []policiesv1.Policy
 
 	clusterAdmissionPolicies, err := f.findClusterAdmissionPoliciesByNamespace(ctx, namespace)
 	if err != nil {
-		return nil, fmt.Errorf("cannot get ClusterAdmissionPolicies: %w", err)
+		return nil, fmt.Errorf("failed to retrieve ClusterAdmissionPolicies for namespace %q: %w", namespace, err)
 	}
 	for _, policy := range clusterAdmissionPolicies {
 		policies = append(policies, &policy)
 	}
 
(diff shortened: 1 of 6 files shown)


internal/policies/client.go
```diff
@@ -90,23 +106,25 @@ func (f *Client) GetClusterWidePolicies(ctx context.Context) (*Policies, error)
...
+		return nil, fmt.Errorf("failed to retrieve cluster-wide ClusterAdmissionPolicyGroups: %w", err)
+	}
+	for _, policy := range clusterAdmissionPolicyGroups {
+		policies = append(policies, &policy)
+	}
+
 	return f.groupPoliciesByGVR(ctx, policies, false)
 }
 
 // findNamespacesForAllClusterAdmissionPolicies returns all the ClusterAdmissionPolicies that evaluate resources in the given namespace.
```
flavio, line 126: nit, you have to change the name of the function also inside of the inline docs. 

Odd, the linter should have spotted it, isn't it?

internal/policies/client.go
```diff
@@ -115,15 +133,34 @@ func (f *Client) findClusterAdmissionPoliciesByNamespace(ctx context.Context, na
...
+
+		if matches {
+			result = append(result, policy)
+		}
+	}
+
+	return result, nil
+}
+
+// findNamespacesForAllClusterAdmissionPolicyGroups returns all the ClusterAdmissionPolicyGroups that evaluate resources in the given namespace.
```
flavio, line 149: Same here, the name should be updated

internal/policies/client.go
```diff
@@ -143,17 +180,56 @@ func (f *Client) listClusterAdmissionPolicies(ctx context.Context) ([]policiesv1
...
+	if err != nil {
+		return nil, fmt.Errorf("cannot list ClusterAdmissionPolicies: %w", err)
+	}
+
+	return clusterAdmissionPolicyGroupList.Items, nil
+}
+
 // listAdmissionPolicies returns all the AdmissionPolicies in the given namespace.
-func (f *Client) listAdmissionPolicies(ctx context.Context, namespace string) ([]policiesv1.AdmissionPolicy, error) {
+func (f *Client) listAdmissionPolicies(ctx context.Context, nsName string) ([]policiesv1.AdmissionPolicy, error) {
```
flavio, line 196: All the `findXXX` functions signatures have been changed by this PR to take a Namespace object instead of its name. Should we be consistent here as well?

flavio (changes requested):
LGTM, I left some minor comments. We can merge the PR once they are solved

PR #377, opened 2024-09-20
test: add a test covering the HTTP error scenario

fabriziosestito:
## Description

 add a test covering the HTTP error scenario


flavio (approved):
LGTM, thanks for the fix

PR #541, opened 2025-06-03
chore(linter): enable wrapcheck linter.

jvanz:
## Description

The wrapcheck linter is disable in the default linter configuration we adopted. But we want to enable it. This commit re enable the linter and fix all the errors it reported.


jvanz:
Waiting for #538 to mark this as ready for review.

PR #620, opened 2025-12-26
chore: replace NewSimpleClientset to NewClientset

pohanhuangtw:
## Description

- replace NewSimpleClientset to NewClientset to fix the linter issue in https://github.com/kubewarden/audit-scanner/pull/619


viccuad (approved):
Many thanks! merging!

